PR #40, opened 2016-04-11
Applications UI

BoykoAlex:
Code has 2 new angular lib dependencies:
1. Angular Bootstrap UI (for modal dialogs)
2. Angular-UI Indeterminate (for indeterminate checkbox)

I noticed there is modal directive in the code base... but I've just used to use Bootstrap UI modal service and this lib has much more goodies we could use in the future... I can try switching to modal directive from SCDF UI code base if necessary though.

### DIFF
--- a/ui/app/scripts/app/dialogs/module-info.html
+++ b/ui/app/scripts/app/dialogs/module-info.html
@@ -4,13 +4,24 @@
     <p ng-if="data.shortDescription">{{data.shortDescription}}</p>
 </div>
 <div class="modal-body" ng-cloak cg-busy="{promise:cgbusy,delay:0,minDuration:100}">
-    <div class="property-group" ng-repeat="property in data.options" ng-if="data.options.length">
-        <label class="col-sm-3 control-label">{{property.name | camelCaseToHuman}}</label>
-        <span class="col-sm-9">
-            <p>{{property.description ? property.description : '(No Description Available)'}}</p>
-            <p class="type-block">{{property.type}}</p>
-        </span>
-    </div>
+    <table class="table table-striped table-hover table-fixed" ng-if="data.options.length">
+        <col width="25%">
+        <col width="75%">
+        <tbody>
+        <tr ng-repeat="property in data.options">
+            <td>
+                <label class="control-label">{{property.name | camelCaseToHuman}}</label>
+            </td>
+            <td>
+                <span>
+                    <p>{{property.description ? property.description : '(No Description Available)'}}</p>
+                    <p class="type-block">{{property.type}}</p>
+                </span>
+            </td>
+        </tr>
+        </tbody>
+    </table>
+
     <div class="row" ng-if="data.options.length === 0">No properties available.</div>
 </div>
 <div class="modal-footer">
--- a/ui/app/scripts/app/dialogs/module-info.html
+++ b/ui/app/scripts/app/dialogs/module-info.html
@@ -1,0 +1,18 @@
+<div class="modal-header">
+    <button type="button" class="close" ng-click="close()">&times;</button>
+    <h4 class="modal-title"><strong>{{name | uppercase}}</strong> ({{type | capitalize}})</h4>
+    <p ng-if="data.shortDescription">{{data.shortDescription}}</p>
+</div>
+<div class="modal-body" ng-cloak cg-busy="{promise:cgbusy,delay:0,minDuration:100}">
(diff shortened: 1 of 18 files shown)


ghillert:
I think I am hesitantly ok with relying on `Angular Bootstrap UI`. Just wanted to point out though, that not relying on it was originally a deliberate decision. This may not be a problem with its increased popularity but at the beginning there were issues where `Angular Bootstrap UI` was not updated to `Bootstrap 3`, yet. Just pointing this out.

A similar situation may play out with `Angular2`, last I checked there is a major impedance with 3rd party directives in regards to considering migrating to `Angular2` later this year. Just something to be aware of.

Also, generally I try to avoid modal dialogs, except for super-simple confirmation dialogs. Instead I favor creating dedicated pages for more complex confirmation or launch screens. That way they can fully partake in the state management of Angular UI and are fully bookmark-able and the back button still works. That way I never had the need for `Angular Bootstrap UI`. Similarly, I solved the tab navigation using pure Angular JS and Angular UI Router (That was the original reason why I had looked into `Angular Bootstrap UI`).

Just my 2 cents.

BoykoAlex:
First of all, thanks for reviewing the PR and feedback :-)

I had a feeling you've looked at the bootstrap-ui lib before. As far as I remember i had a problem with SCDF UI modal directive in the past while working on Flo flavour of SCDF UI... It was something about closing a dialog. Can't remember what it was exactly, I'd have to try re-implement dialogs we have in Flo with modal directive from SCDF UI to refresh my memory. Think it might have been a page with 2 buttons, where each brings up a corresponding modal dialog, I could close one dialog but not the other and had to do something ugly looking up elements by id... When I tried bootstrap-ui there was $modalInstance service that did the trick. So felt like something like $modalInstance was missing. (I admit I might have overlooked something or haven't explored enough)

I saw that they are working on bootstrap-ui based on angular 2 and bootstrap 4, so we might be ok when we switch to angular 2. (https://github.com/ng-bootstrap/core) I didn't see modal support there yet though.

Interesting point about modal dialogs in the UI... We were thinking about possibly submitting another PR in app registration UI area with more complicated dialog... Flo uses dialogs quite a bit too, but don't think we have much choice there.

ui/app/scripts/app/views/register-modules.html
```diff
@@ -0,0 +1,64 @@
+<h1>Register Applications</h1>
+<p>Register one or more applications by entering <strong>Name</strong>, <strong>Type</strong> and <strong>URI</strong> of the application JAR. The URI is typically Maven coordinates of the JAR, but can be local file or docker image URI.</p>
```
ghillert: Wording:
Register one or more applications by entering *a* Name, Type and URI of the application *Jar*. The URI is typically *provided using the* Maven coordinates of the *Jar*, but can be a local file or docker image URI.

We should also provide at least an example for the Maven coordinates, e.g.:

`maven://io.spring.cloud:scdf-batch-simple:jar:exec:1.0.0.BUILD-SNAPSHOT`

ui/app/scripts/app/views/register-modules.html
```diff
@@ -0,0 +1,64 @@
...
+        <col width="7%">
+        <thead>
+            <tr>
+                <th>Name</th>
+                <th>Type</th>
+                <th>URI</th>
+                <th>
+                    <span>
+                        <input type="checkbox" ng-model="toggleAllForce" ng-model-options="{ getterSetter: true }" ui-indeterminate="someForceButNotAll()"/>
+                        Force
```
ghillert: We should provide some pop-over tooltip here, explaining what `force` is.
```
By checking "force", the application will be installed even if it already exists but only if not being used already.
```

ui/app/scripts/app/dialogs/module-info.html
```diff
@@ -0,0 +1,18 @@
+<div class="modal-header">
```
ghillert: Style it similarly to:

http://0.0.0.0:8000/#/tasks/modules/timestamp

So we have the striping in place.

ghillert:
Was going to merge - but found just a few minor things...Once addressed, I can merge.

BoykoAlex:
Looks like wording, popover for 'force' and stripe styling for module details table are all in place.

ghillert:
LGTM - merging

ghillert:
Merged with d9b194a81b1386d9bb3438612a8808075ad3e125

Issue #53, opened 2016-05-19
Problem creating streams in UI when using destinations


aclement:
These streams can be built just fine in the UI but upon pressing the `create stream` button there is some code that incorrectly rebuilds the stream text that is going to be used for the actual definitions. To see the problem open the UI and enter this block of streams:

    :foo > log
    http > :foo
    :a > transform > :b

If you then press the create stream button it will show some proposed stream definitions in the popup dialog, each of the above will be missing a ':' or a '>'

PR #54, opened 2016-05-19
fix variations of create stream with destinations

aclement:
Resolves spring-cloud/spring-cloud-dataflow-ui#53

The fixes are:
- when processing a source channel, if it isn't a tap it will need a leading colon:  :foo > log
- when processing a sink channel, ensure a leading colon is added:  http > :bar
- if using a single module in a bridge like way (but not bridge): `:foo > transform > :bar` ensure enough '>' are added.

Can confirm the fix by attempting to define streams listed in the github issue.


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 6acf817

Issue #59, opened 2016-05-26
Remove unused code


sabbyanandan:
As a developer, I'd like to remove unused code and refactor old terminology references. There are still several references to [xd](https://github.com/spring-cloud/spring-cloud-dataflow-ui/search?utf8=%E2%9C%93&q=xd), [admin](https://github.com/spring-cloud/spring-cloud-dataflow-ui/search?utf8=%E2%9C%93&q=admin), and [containers](https://github.com/spring-cloud/spring-cloud-dataflow-ui/search?utf8=%E2%9C%93&q=container).

aclement:
PR merged.

PR #66, opened 2016-06-03
gh-59 Remove unused code

ghillert:
* Remove references to XD and replace with `dataflow`
* Remove Admin references and replace with `Dashboard`

Resolves spring-cloud/spring-cloud-dataflow-ui#59


ghillert:
Fixed E2E tests. All should be good to go now.

ericbottard:
I'm getting the following error after building with maven and trying to run dataflow local (using the newly created ui artifact):

```
angular.js:63 Uncaught Error: [$injector:modulerr] Failed to instantiate module dataflowMain due to:
Error: [$injector:nomod] Module 'dataflowMain' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
```

aclement:
This clashes with a couple of Alexs PRs, I'm making the necessary adjustments.

aclement:
Merged and pushed. Minor tweaks:
- had to refactor the code Alex introduced to handle duplicate stream names in flo to take account of the renaming of various services (followed the new naming scheme)
- modified the e2e tests again. We now have more than just timestamp on the tasks apps view so changed the tests to look for timestamp before interacting with it rather than assuming it is the first one on the list.
- in the spirit of removing xd references, modified the deploy stream page which had an XD doc reference to have a dataflow doc reference.

I did not see the error Eric showed above, not sure why. All running well for me (standalone via grunt serve and after packaging into a real dataflow local server build)

PR #76, opened 2016-06-11
Enable UI sections based on feature enablement info from the server

### LINKED ISSUE #64: Support features endpoint to disable sections of the UI
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/640

If the task support is disabled, the UI should not show task/job related components.

BoykoAlex:
Resolves #64

### DIFF
--- a/ui/app/index.html
+++ b/ui/app/index.html
@@ -46,16 +46,16 @@
             <li ng-class="{ active: $state.includes('home.runtime') }">
                <a ui-sref="home.runtime.tabs.runtimeAppsList">Runtime</a>
             </li>
-            <li ng-class="{ active: $state.includes('home.streams') }">
+            <li ng-show="features.streamsEnabled" ng-class="{ active: $state.includes('home.streams') }">
                <a ui-sref="home.streams.tabs.definitions">Streams</a>
             </li>
-            <li ng-class="{ active: $state.includes('home.tasks') }">
+            <li ng-show="features.tasksEnabled" ng-class="{ active: $state.includes('home.tasks') }">
               <a ui-sref="home.tasks.tabs.definitions">Tasks</a>
             </li>
-            <li ng-class="{ active: $state.includes('home.jobs') }">
+            <li ng-show="features.tasksEnabled" ng-class="{ active: $state.includes('home.jobs') }">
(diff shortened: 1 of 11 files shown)


BoykoAlex:
Took minimal amount of changes direction.
Let me know if it's best to follow `securityInfo` constant pattern for features enablement. That was my first solution, but blocking the UI until `/features` request is resolved seemed like not a good idea.

ilayaperumalg:
Given the `dashboard` home page always redirects to the route `/streams/definitions`, we either need a different route (possibly `/apps`) or make the `features` info available before routes (which makes it look like having constant like `securityInfo`).

ilayaperumalg:
We need to decide if we are ok with:
1) The `/apps` route can be the default in the UI page (since that feature is always available)
2) Fix the redirection based on the features condition as mentioned above.

sabbyanandan:
Let's please default to `/apps` as the landing page.

BoykoAlex:
I have made `/apps/apps` as the default URL

BoykoAlex:
I have added the fix for direct navigation to `/streams` `/analytics`, `/tasks` and `/jobs`. Was rather trivial... A bit more complication on reviewing `routes.js`, but i think changes are obvious and easy to follow.

ui/app/index.html
```diff
@@ -46,16 +46,16 @@
             <li ng-class="{ active: $state.includes('home.runtime') }">
                <a ui-sref="home.runtime.tabs.runtimeAppsList">Runtime</a>
             </li>
-            <li ng-class="{ active: $state.includes('home.streams') }">
+            <li ng-show="features.streamsEnabled" ng-class="{ active: $state.includes('home.streams') }">
```
ilayaperumalg: The `navbar-brand` needs to be fixed:

> <a class="navbar-brand" ui-sref="home.jobs.tabs.definitions"><span></span></a>

will fix while merging.

ilayaperumalg:
@BoykoAlex Thank you for the quick turnaround. The changes look great. merging.

ilayaperumalg:
Squashed, rebased and merged as 3e7aa95237c888bde4c6ec8f8f32a230562b9ece

PR #85, opened 2016-06-14
Adopt whitelisting

BoykoAlex:
To merge once SCDF and app-starters related PRs are merged

### DIFF
--- a/ui/app/scripts/stream/services/metamodel.js
+++ b/ui/app/scripts/stream/services/metamodel.js
@@ -26,22 +26,6 @@ define(function (require) {
 
     var keepAllProperties = false; // TODO address when whitelisting properties fixed in SCDF
 
-    var WHITELISTING_NOT_SUPPORTED_APPS = [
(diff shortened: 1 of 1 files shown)


sabbyanandan:
Depends on spring-cloud/spring-cloud-stream-app-starters#127

aclement:
Looks ok - if I were testing I'd try a whitelisted app, a non whitelisted app and the scriptable-transform-processor which has special handling for the script field. Check:
- the properties view looks reasonable in each case when opened for these elements
- after entering data into the properties view the text view uses the right names for the properties
- modify the textbox and check when you open the properties view the changes are reflected
- if there are aliases for properties (long/shot name - are there aliases?) verify what happens if using both names in the textbox - what does the properties view do?

After defining some samples including whitelisted/non-whitelisted properties actually deploy them and verify they deploy ok.

sabbyanandan:
This has been already merged.

Issue #75, opened 2016-06-10
Problem while creating new stream from UI in IE11


sachkk76:
I was trying to patch spring-cloud-dataflow-ui-1.0.0.M2 version with spring-cloud-dataflow-server-cloudfoundry-1.0.0.M2,,as per instructions at  https://docs.pivotal.io/spring-cloud-data-flow/installing-scdf.html. but did not work.
Then I had patched spring-cloud-dataflow-ui-1.0.0.M3 with  spring-cloud-dataflow-server-cloudfoundry-1.0.0.M2.. and it worked.

So now flo UI up and running I see few problems -
1) Creating new stream from UI - when you drag drop time source and log sink to create tick tock stream...create stream  pop-up window stays there doing nothing.. no logs on dataflow-server
Same stream I can create from dataflow-shell.
Is there any issue patching flo with dataflow-server-cloudfoundry-1.0.0.M2?

2) New stream created (from shell) does not get reflected in runtime or stream definition unless browser cache is cleaned.

BoykoAlex:
1) Is the create stream dialog staying up after you've clicked on Create button on the dialog?
2) Doesn't sound right. If you go to "Stream" tab and then open "Definitions" sub-tab you should streams created from the shell there. The list is refreshed every 2 secs - no need to wipe out the cache...

sachkk76:
yeah "Confirm stream(s) Creation" dialog stays up when I clicked on create button forever..

sabbyanandan:
@sachkk76: 

> yeah "Confirm stream(s) Creation" dialog stays up when I clicked on create button forever..

I'm assuming you hit "Create" button from dialog and the window just hung there - is that correct? Please note that you're deploying the stream to PCF. It is not instantaneous. In other words, all the stream apps needs deployed successfully to PCF - it takes 1-2 mins for individual apps to go through `cf push` lifecycle. With that said, yes, we would need to add a progress bar to indicate the progress in the UI.

> New stream created (from shell) does not get reflected in runtime or stream definition unless browser cache is cleaned.

Browser cache has nothing to do with stream deployments though. Could you elaborate on what you mean by this?

sabbyanandan:
And, oh.. 

> I was trying to patch spring-cloud-dataflow-ui-1.0.0.M2 version with spring-cloud-dataflow-server-cloudfoundry-1.0.0.M2

[...]

PR #86, opened 2016-06-16
IE 11, Edge, old browser versions don't support Array#findIndex()

BoykoAlex:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/75
Should also resolve the same issue for Chrome version <45

There are other UI glitches on IE11 and Edge... Tooltips aren't removed under some conditions... Zooming slider is ugly... This is just for the dialog to close because it's a real blocker

### DIFF
--- a/ui/app/scripts/stream/controllers/create-streams-dialog.js
+++ b/ui/app/scripts/stream/controllers/create-streams-dialog.js
@@ -106,9 +106,11 @@ define(function(require) {
                     });
                 } else {
                     // Find index of the first not yet created stream
-                    var index = $scope.streamdefs.findIndex(function(def) {
-                        return !def.created;
-                    });
+                    // Can't use Array#findIndex(...) because not all browsers support it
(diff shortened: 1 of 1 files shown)


markpollack:
I took this one since I have the browser in question that demonstrated this issue.

markpollack:
This PR did not solve the problem for me.  It also raised the question of determining which UI build I was using.  The 'about' tab shows the server version but not the UI version - that would help make sure it wasn't a build error.  I built the UI locally via `mvn clean package install` and then a local build of the dataflow project.

markpollack:
realized the pom for dataflow was pulling in 1.0 RC1 of the ui...retesting.

ilayaperumalg:
Yeah, we haven't switched back to `SNAPSHOT` for the UI after RC1.

markpollack:
Verified it works on chrome 43.0.2357.132 (64-bit) linux

PR #89, opened 2016-06-22
Support either long-names or short-names for properties

aclement:
Prior to this change Flo only recognized use of the short names
and any use of long names would not get reflected in the
properties view.

With this change the DSL text can include either short or
long names and the properties view will correctly pickup any
specified values. When the properties view is closed the
DSL text will be updated to reflect new values and should use
the same form of name (long or short) that was originally
used. If the properties view is used to set a property
that previously had no value the DSL text will use the
short name.

### DIFF
--- a/ui/app/scripts/stream/controllers/properties-dialog.js
+++ b/ui/app/scripts/stream/controllers/properties-dialog.js
@@ -75,9 +75,25 @@ define([], function () {
                 Object.keys(schemaProperties).forEach(function (key) {
                     schema = schemaProperties[key];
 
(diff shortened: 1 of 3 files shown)


ui/app/scripts/stream/controllers/properties-dialog.js
```diff
@@ -99,7 +115,7 @@ define([], function () {
                             }
                         }
                     };
-
+
```
ghillert: remove white-space

ghillert:
* Fixed white-space issue

Merged with commit dd66d5c.

PR #90, opened 2016-06-24
Temporarily deactivate eager stream name validation in create tab

aclement:
When building streams it is nice to eagerly validate stream
names and let the user know if they are specifying something
invalid as early as possible.  Unfortunately the current
endpoint logs an exception on the server when a check is made
for a name that does not currently exist. Until we can change
that to avoid the exception for our scenario, this commit
turns off validation.

I did not delete the code because we want to turn it back on as soon
as we can tweak the endpoint on the server. But if you really
want me to, I can do that. For similar reasons I did not 'undo' all the
validation infrastructure that has been built around this feature, I wanted
to keep this change minimal. Comments welcome.

### DIFF
--- a/ui/app/scripts/stream/directives.js
+++ b/ui/app/scripts/stream/directives.js
@@ -36,20 +36,26 @@ define(function(require) {
                             return utils.$q.when();
                         }
 
-                        var def = utils.$q.defer();
-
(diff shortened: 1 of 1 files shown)


ghillert:
Merged with commit dd66d5c.

Removed some trailing white-space.

PR #92, opened 2016-06-29
Fix e2e tests

ghillert:


aclement:
LGTM, merging.

markpollack:
FWIW, the controllers now log this as a warning without a stack trace, the exception message is logged at WARN level unless TRACE logging level is enabled.

PR #96, opened 2016-07-07
Fix Scriptable-Transform regression, make hard-coded metamodel data more robust

BoykoAlex:


ui/app/scripts/stream/services/metamodel.js
```diff
@@ -183,7 +183,14 @@ define(function (require) {
                                     } else if (p.sourceType === 'org.springframework.cloud.stream.app.scriptable.transform.processor.ScriptableTransformProcessorProperties') {
                                         if (p.name === 'script') {
                                             p.contentType = 'Plain Text';
-                                            p.contentTypeProperty = 'language';
+                                            // Add proeprty prefix if any is found.
```
ghillert: `proeprty` --> `property`

ghillert:
@BoykoAlex LGTM - merging. Will fix spelling error during merge. If we a chance to maybe target a TypeScript/Angular2 re-write later this year, we must encapsulate as much of the code base in tests as possible. We start to see more and more logic in the JS code base, which makes me nervous.

PR #100, opened 2016-07-11
Adopt scriptable-transform tooltip to properties prefix

BoykoAlex:


ghillert:
Merged with commit e988396

PR #105, opened 2016-08-09
Read-only stream defs on Stream Definitions page

BoykoAlex:


ghillert:
Fixed some trailing white-space. Merged with commit 8ea4bee.

Issue #58, opened 2016-05-24
Dashboard won't load


sabbyanandan:
As a user, I'm trying to launch the dashboard; however, I only see the `loading...` prompt and the SPA never renders. Browser console includes the attached `require.js` timeout error. Similar errors with Local and PCF dataflow-server variants. 

Note: This is noticed sporadically and it is not consistently repeatable. Perhaps there's some default timeout setting that needs bumped to allow backend service calls to complete in reasonable time?

![image](https://cloud.githubusercontent.com/assets/1562654/15505424/5eea9f9a-2178-11e6-89c0-38345720210b.png)

BoykoAlex:
Default is 7 but we can indeed bump the value (http://requirejs.org/docs/api.html#config-waitSeconds)

ghillert:
It was not my decision / desire to ever use requirejs - just for the record :-).

markpollack:
Yes, but is that the root cause or some crap network and a bad timeout value?  Would there be no timeout to download angular.js w/o require.js.  Also, we could revisit the use of require.js if it seems to be an issue for us going forward.  Certainly we are not making use of it's feature set ATM.

PR #109, opened 2016-08-12
Specified Require JS script load timeout

BoykoAlex:
Should fix https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/58


ghillert:
Merged with commit 49aa5e9 (minor white-space trim during merge)

PR #110, opened 2016-08-23
Tooltip rendering

BoykoAlex:
Fixing the following bug with the canvas shape tooltip: https://www.pivotaltracker.com/story/show/127194631
Also requires Flo Core fix, consequently switching Flo Core dependency to repository master branch until the release.


ghillert:
Merged with commit 0af2210

PR #111, opened 2016-08-23
Create stream progress

BoykoAlex:
Progress bar for create stream dialog.


ghillert:
@BoykoAlex Sorry for being so late to the party. Question - what is the reason to use a progress bar indicator? Rather than using the "Busy indicator" which we use across all other screens?

The only other place, where we have something remotely similar is the progress bar for Batch (Step execution progress bar).

I see it is for the use-case where you create multiple streams at once, for which use-case it makes sense. But most user would create one stream only most of the time, no?

Just wondering from a UX perspective. For just 1 stream, the progress bar has no value. It will never show anything

Also, if we stick with the progress - we should have it shown for `500ms-1s` otherwise it is just like a flicker when the backend is too fast. Ideally it should have something like the following sequence:

* Show progress-bar at `0%` for at least `500ms`
* Show progress-bar at `100%` for at least `500ms`

We may play with the numbers a bit but I think this would create a more fluid meaningful user experience. 

Lastly, we should problem show a success message up successful creation of streams. Right now there is no success message. Did my stream get created?

BoykoAlex:
I could have used `cg-busy` as before... Multiple streams creation may not be covered well with `cg-busy` overlay: stream creation request is sent then waiting until new stream resource becomes available then the next stream and so on until the last stream... Progress bar seem to be a better fit. Also, frankly, I was just eager to try `angular-bootstrap-ui` progress bar directive :-)

Good point on the success growl message on successful stream(s) creation operation. I'll put that in.

I'll try to animate progress-bar a bit differently such that it shows up for at least a second for the fast stream(s) creation.

BoykoAlex:
@ghillert lets try again :-) I added growl message on success and ensured progress bar stays up for a bit for fast create stream operation.

ghillert:
@BoykoAlex Looks almost perfect - I just need one more change. Using the timeout is too simplistic as it means to *always* wait 500ms + execution time.

The logic should be instead:
```
if (executionTime < 500ms) {
   wait(500ms - executionTime)
}
else {
  neverWait()
}
```

BoykoAlex:
There is animation on the progress bar. Therefore, once stream creation has completed the progress bar is set to 100% and it'll take some ~0.5s to animate the bar to reach 100%. Thus, I'd wait in both cases when stream creation has completed to let the progress bar reach 100%. Then if we are to wait in both cases for the progress bar to get to 100% then the 0.5s simple wait time seems ok, isn't it?

ghillert:
@BoykoAlex The problem I mean is that the REST Api call does not start until after the wait-time completes. I think the REST call should trigger right after pressing the "create stream" button.

BoykoAlex:
@ghillert I have removed the wait before the REST call, hence it's made right away. There is a wait time when stream(s) creation operation has completed. The wait time allows for the animated progress bar to get to 100% as well as ensure that progress bar is visible to a user in the case of the fast operation completion. Hope I understood you correctly...

ghillert:
Looks much better :-) Merged with commit 7101ef089f222c3f3751ba66b722c2f5e9c0ee7a

ghillert:
Also back-ported to https://github.com/spring-cloud/spring-cloud-dataflow-ui/tree/1.0.x

PR #121, opened 2016-09-20
Bulk define tasks with validation

BoykoAlex:


aclement:
Just a quick description of the changes:
- There is a new bulk define tasks action on the tasks>definitions page which opens a new dialog for entering task definitions
- The parser service, previously only used for streams has been moved to the 'shared' area because it is now also used for task parsing too
- there is a new controller for the bulk define tasks page. The page is basically one big text box.
- when data is entered the new validation service is called to check the syntax of the text *and* whether the tasks named in the textbox actually exist on the server.
- The validation service feeds back errors to the dialog page (asynchronously) which indicates how many problems there are and by clicking the errors count or scrolling the textbox you can quickly find them (they are highlighted in the textbox).
- There are a bunch of tests for the new uses of the parser (for parsing tasks) and the validation service.

BoykoAlex:
DO NOT LOOK AT THIS PR.
REVIEW THIS ONE INSTEAD: https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/126

PR #129, opened 2016-10-06
Syntax validation only for tasks without app name/properties verification

BoykoAlex:
- Support for turning on/off semantic validation of app and their options
- UI should have button "Verify Apps" added on the bulk define tasks page
- Unit tests for the option for the validator

### DIFF
--- a/ui/app/scripts/task/app.js
+++ b/ui/app/scripts/task/app.js
@@ -29,6 +29,7 @@ define([
   'ngAnimate',
   'pagination',
   'angularHighlightjs',
+  'ngCookies',
   './controllers',
   './services',
   '../directives',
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Given we only do DSL syntax verification when `verifyApps` is **not** invoked, can we specify the validation result to say *only* about `syntax` instead of saying `valid task definition detected`?

BoykoAlex:
Yes. Changed to `syntactically correct` when `verifyApps` flag is off

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 7ffcc28c5c176ec5947fec407741e91913390273

PR #130, opened 2016-10-06
Clear input file value after file is uploaded

BoykoAlex:
Reproduce:
1. Open "Bulk Define Tasks" page
2. Import file XYZ.txt
3. Select all contents of the editor and remove them, i.e. editor content is empty.
4. Import file XYZ.txt again

Note that content of the editor is still empty. Happened because file path value is still the same and there is nothing to upload.

Solution: added directive clearing file path value from the DOM input element

See: https://www.pivotaltracker.com/story/show/131820453

### DIFF
--- a/ui/app/scripts/directives.js
+++ b/ui/app/scripts/directives.js
@@ -382,5 +382,16 @@ define(['angular', 'xregexp', 'moment'], function(angular) {
           });
         }
       };
+    })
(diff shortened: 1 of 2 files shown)


ghillert:
Verified issue - Looks good. Merging.

ghillert:
Merged with commit 8da6474669f2c51b904ada31682754f506cbac10

PR #131, opened 2016-10-06
Fix reference from StreamParserService to shared ParserService

aclement:
Without this I think the create button on the Create Streams page is broken (please verify that before you apply). It will open the dialog but then fail because it is looking for the streams parser (when it is now a shared parser). With this change in place it will find the right thing and succeed.

We don't have enough automated tests around that UI.


BoykoAlex:
LGTM. Merged.

PR #142, opened 2016-11-07
Use the new backend support for content assist in bulk task define

aclement:
1. Introduce a new service for invoking the /completions/task backend endpoint
2. Use that new service in the bulk-define-tasks UI

There is some mucking about to trim the proposals such that they line up nicely (the default proposals from the backend are very long).


BoykoAlex:
Merged with b5c44dbd4fd1d3facbb375156f51ec47169cb3ac. Attempted to fix some trailing white spaces.

PR #157, opened 2017-02-14
Quick fix for download links

trisberg:
- pointing shell download to 'libs-snapshot' repo to resolve release, milestone amd snapshot artifacts

- updating app resources to use Avogadto.SR1 and Addison.GA


ghillert:
LGTM - merging.

Issue #158, opened 2017-02-15
Dashboard refresh causes user session logout


sabbyanandan:
As a user, I'm using the file-based authentication and simple authorization rules as attached below. After successful login, I'm able to access dashboard according to the defined permissions; however, when I reload the dashboard, I lose by login session and I'm redirected to login page instead.

```
security:
  basic:
    enabled: true
    realm: Spring Cloud Data Flow
spring:
  cloud:
    dataflow:
      security:
        authentication:
          file:
            enabled: true                                                 
            users:                                                        
              bob: bob, ROLE_MANAGE
              alice: alice, ROLE_VIEW, ROLE_CREATE
```

PR #163, opened 2017-02-27
gh-158 Fix: Dashboard refresh causes user session logout

ghillert:
* Store oauth-token in session-store
* Restore Dashboard UI in-case a pre-existing oauth-token can be retrieved AND if the SecurityInfo REST endpoint indicates that the user is still logged in
* Polish login page
  - Improve error handling
  - Disable login button as long as both form-fields (username / password) are not filled
* Logout will clear the session-store

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/158

### DIFF
--- a/ui/app/scripts/auth/controllers/login.js
+++ b/ui/app/scripts/auth/controllers/login.js
@@ -23,37 +23,50 @@
 
 define([], function () {
   'use strict';
-  return ['$scope', '$state', 'userService', 'DataflowUtils', '$log', '$rootScope', '$http',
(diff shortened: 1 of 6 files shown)


BoykoAlex:
LGTM. Just a note: `userService#restoreUser(xAuthToken)` doesn't seem to be used anywhere...
Merging.

BoykoAlex:
Merged with b2894499644bb92152a75165ed73fac6b61aa48d

Issue #165, opened 2017-02-28
Remove bulk-registration of apps from the UI


sabbyanandan:
As a developer, I'd like to remove the bulk registration of apps from the UI in entirety. 

Reasons:
- creates strong coupling between app-starter releases with SCDF-server
- deviates from SCDF's core design premise (eg., register and use the apps  _only_ when needed as opposed to registering them all)
- pose a confusion for users wrongly registering Docker artifacts on the runtime platforms where it isn't compatible/supported

PR #166, opened 2017-03-02
gh-165 Remove bulk-registration of apps from the UI

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/165

### DIFF
--- a/ui/app/scripts/app/controllers/bulk-import-apps.js
+++ b/ui/app/scripts/app/controllers/bulk-import-apps.js
@@ -22,9 +22,6 @@
 define(function () {
     'use strict';
 
-    var STREAM_APP_STARTERS_VERSION_SUFFIX = 'Avogadro-SR1-';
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
From #165 I assumed we wanted to remove the `bulk import` page itself. From the current changes on this PR, I see only the OOB app starters list for stream/task apps are removed from the `bulk import` page. @sabbyanandan @ghillert Could you clarify?

ghillert:
@sabbyanandan Maybe you could just past your slack answer to my question here. Just double-checking that I have understood the requirements correctly.

sabbyanandan:
@ilayaperumalg: The text-area is still applicable if someone wants to copy+paste app coordinates and register them that way. I don't see it confusing compared with how we had through bit.ly's previously. If there are any other concerns, it is fine to remove that as well.

ilayaperumalg:
@sabbyanandan @ghillert I think it makes sense to keep the `bulk import` page at least if this wasn't reported as a confusion. LGTM, merging.

Issue #180, opened 2017-03-27
Convert stream deployment property textbox to k/v pairs


sabbyanandan:
As a user, I'd like to pass deployment properties as k/v pairs, so it is easy to verify all properties as individual string literals as opposed to a comma separated string in textbox. 

**Acceptance:**
- Convert deployment property textbox to k/v pairs
- Render k/v pairs similar to task launch properties experience
- Parsing of k/v pairs with backend is handled appropriately
- Update screenshots in the reference guide

Issue #175, opened 2017-03-20
Add support for --propertiesFile equivalent from the stream deploy view


sabbyanandan:
As a user, I'd like to have the option to supply stream deployment properties via properties file.

**Acceptance:**
- stream deployment command accepts an optional file upload capability
- only properties files is accepted

PR #181, opened 2017-03-28
gh-175 Add support for `--propertiesFile` equivalent from the stream deploy view

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/175
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/180

### DIFF
--- a/ui/app/scripts/stream/controllers/definition-deploy.js
+++ b/ui/app/scripts/stream/controllers/definition-deploy.js
@@ -49,6 +49,12 @@ define([], function () {
         utils.$log.info('Canceling Stream Definition Deployment');
         $state.go('home.streams.tabs.definitions');
       };
+      $scope.displayFileContents = function(contents) {
+        console.log(contents);
+        $scope.$watch('definitionDeployRequest', function () {
+          $scope.definitionDeployRequest.deploymentProperties = contents.split('\n');
+        });
+      };
       $scope.deployDefinition = function (definitionDeployRequest) {
         utils.$log.info('Deploying Stream Definition ' + definitionDeployRequest.streamDefinition.name, definitionDeployRequest);
         var properties = definitionDeployRequest.deploymentProperties;
--- a/ui/app/scripts/stream/controllers/definition-deploy.js
+++ b/ui/app/scripts/stream/controllers/definition-deploy.js
@@ -59,9 +59,24 @@ define([], function () {
         utils.$log.info('Deploying Stream Definition ' + definitionDeployRequest.streamDefinition.name, definitionDeployRequest);
         var properties = definitionDeployRequest.deploymentProperties;
 
-        utils.$log.info('Deployment Properties:', properties);
+        var propertiesAsMap = null;
 
-        streamService.deploy(definitionDeployRequest.streamDefinition, properties).$promise.then(
+        if (properties) {
+          propertiesAsMap = {};
+          for (let prop of properties) {
+            var keyValue = prop.split('=');
+            if (keyValue.length===2) {
+              propertiesAsMap[keyValue[0]] = keyValue[1];
+            }
+            else {
+              utils.$log.warn('Invalid deployment property "' + prop +'" must contain a single "=".');
+              $scope.deployDefinitionForm.deploymentProperties.$setValidity('invalidDeploymentProperties', false);
+              return;
+            }
+          }
+        }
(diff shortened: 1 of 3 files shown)


ericbottard:
Seems there was some misunderstanding

This change should be coupled with https://github.com/spring-cloud/spring-cloud-dataflow/pull/1229 (ie post JSON data).
Also, I'm not sure I understand how this solves the problem of parsing on the client side. Until we have *dedicated* fields for the key and the value, we'll always have the need to parse on the client side. Making the box bigger doesn't change anything (although it is more readable :) )

ilayaperumalg:
@ericbottard Not sure you mean to refer https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/180. 

But, I agree on the client side parsing side of issues here. For example, the `--propertiesFile` also supports sending YAML file and hence, we need to take care of converting YAML properties as well.

ghillert:
@ericbottard You reference https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/180 - which is different (If we only consider `Properties`). Actually, with this PR, #180 can be supported with literally a 1-line change. We do that same thing for the bulk import of apps.

**However, if we need Yaml support**

@sabbyanandan I did not implement `yaml` support as the issue description restricted it to `only properties files are accepted`. In order to support `yaml` or maybe even `json`, we should tackle #180 but should broaden its scope. 

@ericbottard @sabbyanandan @ilayaperumalg In that case we should maybe even explicitly support the different property formats for the REST endpoint directly, so that clients can be `simpler`:

* yaml
* properties (comma separated [old style], new-line separated [translates to `Properties` on the server-side])
* json

**What should be the default?**

While there are `yaml` parsers for JS, it adds some complexity and possibly minor differences in behavior if we use different libraries for parsing in the Shell and browser clients. Thus, the idea of having the parsing done server-side.

ericbottard:
The REST endpoint doesn't mind about the property format. It's content negotiated (I said JSON earlier but that was actually a bad choice of terms to mean "key value pairs", sorry). As long as there is an `HttpMessageConverter` that supports it, we're fine. This could be JSON or form-encoding.

The point is, we need the values to be actual key/value pairs in the payload, as opposed to be a simple String (that is then ill-parsed on the server side) under the key `properties` as it is now.

If there is some kind of "format support", then that happens entirely on the client side.

ghillert:
@ericbottard 

In the Shell we use `YamlPropertiesFactoryBean` ([see here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java#L130))

to parse Yaml files being specified via `--propertiesFile`. In the browser, at present time, I would need to use something like [JS-YAML](https://github.com/nodeca/js-yaml) but:

```
We have no resources to support browserified version. Don't expect it to be well tested. Don't expect fast fixes if something goes wrong there.
```

Plus there is no guarantee that **Yaml** parsing via `JS-YAML` differs from the Java version. Therefore, I would favor some approach where your PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/1229 would also support Yaml directly.

sabbyanandan:
After discussing this with @markpollack and @ghillert, we will be supporting _only_ ".properties" file as first step. The YAML support will be added post 1.2 release.

jvalkeal:
@ghillert I merged spring-cloud/spring-cloud-dataflow/pull/1229 so you can now fix this PR

ghillert:
@jvalkeal Updated PR, also added some basic form validation.

jvalkeal:
works ok, merging

ilayaperumalg:
@jvalkeal Please hold on merging this PR. I see some issues with handling the properties file with empty line or with new line at EOF. I have a fix for it and testing it now.

ilayaperumalg:
@jvalkeal Here is the gist of the changes I have made and tested ok: https://gist.github.com/ilayaperumalg/ef37276e39410bb47a2ba75e2cf9ab58

jvalkeal:
Merged/squashed with changes discussed with ilya per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/5cac1b79d48bf3d1db16ceb89b94155b2d36902e

PR #187, opened 2017-04-05
Remove composed job/task flag from execution detail page

### LINKED ISSUE #186: Remove composed job/task flag from execution detail page
Since there is no info on composed task/job in task/job execution detail,
we can remove the html content to avoid confusion

ilayaperumalg:
- Since there is no info on composed task/job in task/job execution detail,
we can remove the html content to avoid confusion

Resolves spring-cloud/spring-cloud-dataflow-ui#186


BoykoAlex:
LGTM. Merged with d17bd76b76f8cb013e721aa2dd47d4d3b78a00ff

Issue #170, opened 2017-03-12
Add Flo support for composed tasks


sabbyanandan:
As a user, I'd like to have a visual canvas to compose tasks, so I can define and create task topologies graphically.

**Acceptance:**
- A new Task canvas is available
- Canvas includes both drag&drop and DSL capabilities
- Includes Graph to Text and Text to Graph capabilities
- Action based validation rules evaluated in real-time

PR #188, opened 2017-04-05
Composed Tasks

BoykoAlex:
Composed Tasks UI reanimated from Flo for Spring XD

Resolves spring-cloud/spring-cloud-dataflow-ui#170


ui/app/scripts/task/controllers/definitions.js
```diff
@@ -90,6 +118,31 @@ define(['model/pageable'], function (Pageable) {
       $scope.bulkDefineTasks = function() {
         $state.go('home.tasks.bulkDefineTasks');
       };
+      $scope.isComposedTask = function(item) {
+        return item && angular.isString(item.dslText) && item.dslText.startsWith('composed-task-runner ');
```
ilayaperumalg: The name `composed-task-runner` isn't a fixed name for all the composed task and one can define custom name for this via dataflow property. Hence, we need a better approach to decide if the task is composed or not.

ilayaperumalg: May be we can look for `--graph` and probably some syntax that is applicable only for the composed task?

ui/app/scripts/app/controllers/apps.js
```diff
@@ -81,7 +81,7 @@ define(['model/pageable'], function (Pageable) {
 
                                 // Define function for obtaining properties of a app
                                 item.getProperties = function() {
-                                    return appService.getAppInfo(this.type, this.name);
+                                    return appService.getAppInfo(this.type, this.name).$promise;
```
ilayaperumalg: While it definitely makes sense fix this to return Promise, I wonder what made this work fine so far? ;)

ui/app/scripts/app/services.js
```diff
@@ -73,10 +74,15 @@ define(['angular', 'lodash'], function (angular, _) {
...
-                    return $http({
-                        method: 'GET',
-                        url: $rootScope.dataflowServerUrl + '/apps/'+appType+'/'+appName
-                    });
+                    return $resource($rootScope.dataflowServerUrl + '/apps/'+appType+'/'+appName, {}, {
+                        query: {
+                            method: 'GET'
+                        }
+                    }).get();
+                    // return $http({
```
ilayaperumalg: Please remove the commented out lines (on this PR) unless they are really needed for references.

ui/app/scripts/routes.js
```diff
@@ -306,6 +306,24 @@ define(['./app'], function (dashboard) {
...
+      url : 'tasks/definitions/{taskName}',
+      templateUrl : taskTemplatesPath + '/detailed-task.html',
+      controller: 'DetailedTaskController',
+      data:{
+        authenticate: true,
+        feature: 'tasksEnabled'
+      }
+    })
+    .state('home.tasks.tabs.createComposedTask', {
+      url: '/createComposedJob',
```
ilayaperumalg: Is this URL defined as a data flow server REST endpoint already?

ui/app/scripts/app/controllers/apps.js
BoykoAlex: `$http` service returns a promise already. I have switched it to `$resource` service that seems more appropriate and it returns an object with a `$promise` property.

ui/app/scripts/routes.js
BoykoAlex: Nope. The endpoint used to be `tasks/composed-definitions` but it's gone now. The regular task definition creation endpoint `tasks/definitions` creates composed task definition now.

ui/app/scripts/app/services.js
BoykoAlex: Sounds good.

ui/app/scripts/task/controllers/definitions.js
BoykoAlex: Thanks! Good point! I'll look for `--graph` at least... maybe something more...

BoykoAlex:
@ilayaperumalg Updated the PR per your comments. Lts just look for `--graph="XXX"` in the DSL. If `graph` property is there it's fair to show the graph

BoykoAlex:
Squashed commits

ui/app/scripts/app/controllers/apps.js
ilayaperumalg: ok, I see that in `getAppInfo` service now. thanks!

ui/app/scripts/routes.js
ilayaperumalg: ok, In that case, what does this `/createComposedJob` do?

BoykoAlex: It opens up an `Create Composed Task` editor UI tab. I guess it shouldn't be `/createComposedJob`. I have changed it to `/create-composed-task`. It's the same as `/definitions` URL for task definitions tab contents page. The full URL would have `/tasks/` prefix.

trisberg:
When I create a stream I can type in a DSL like:

```
ticktock = time | log
```

Trying something similar with composed tasks:

```
foobar = foo && bar
```

but that's not accepted

Also, when I create a composed task I see the message that task has been created and the graph is erased, could we show the graph or switch to the definitions view instead of staying at a blank Create Composed Task view

BoykoAlex:
@trisberg I think that is because DSL doesn't allow the name to be in the definition. User is allowed to enter the name of the task when creation dialog is shown. Although, I agree that perhaps we should have been consistent with streams I doubt this can be easily fixed now.
Navigating to `definitions` page after creation is debatable... It's not done for streams either. Also we'd have to navigate to the last page of definitions probably to show it... Not sure... In any case I'd rather push all these changes now and then polish and add small features on top of it.

aclement:
I take Thomas' point about supporting the same as streams "foo = aaa && bbb", we should do that but I'm ok with that not being in MVP. Worth tackling when we get into the DSL syntax post this version because we have a lot to discuss there.

ilayaperumalg:
LGTM,merging.

ilayaperumalg:
Squashed, rebased and merged as fa84467beacdd90bfd855087d6baffd6c061248e

PR #212, opened 2017-04-25
Truncate text with ellipsis tooltip width adjusted based on content

### LINKED ISSUE #204: Flo instance metric popup too tiny
_From @jvalkeal on April 20, 2017 18:44_

On yarn `guid` will be something like `container_1492712105510_0003_01_000003` and I think it is causing `guid` and actual metric value to span outside of popup itself. 
metric-ui-bug-report

Can we fit these a little better?

_Copied from original issue: spring-cloud/spring-cloud-dataflow#1378_

BoykoAlex:
Resolves #204

### DIFF
--- a/ui/app/scripts/stream/controllers/stream.js
+++ b/ui/app/scripts/stream/controllers/stream.js
@@ -429,7 +429,7 @@ define(function (require) {
 
                                     floBootstrapTooltip.attachBootstrapTemplateTooltip(dotView.el, dotScope,
                                         'moduleInstanceData ? \'scripts/stream/views/app-deployed-tooltip.html\' : \'scripts/stream/views/app-undeployed-tooltip.html\'',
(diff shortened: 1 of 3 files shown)


jvalkeal:
Not sure If I messed up something by compiling snapshots into dataflow but flo graph is now totally gone from definitions.
broken-flo

jvalkeal:
bummer, still pulled in ui RC1 so I messed something.

jvalkeal:
there's is a bug in dataflow caused by pagination changes and angular goes to loop. I'll check this again when that bug is fixed.

jvalkeal:
caused by https://github.com/spring-cloud/spring-cloud-dataflow/issues/1397

jvalkeal:
Yes, this PR looks good and popup is now better sized. It probably goes to 1.2.1.

Issue #216, opened 2017-05-12
Change dashboard title


sabbyanandan:
As a developer, I'd like to replace the [dashboard title](https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/b2894499644bb92152a75165ed73fac6b61aa48d/ui/app/scripts/auth/views/login.html#L6) with the following from project site.

> Spring Cloud Data Flow is a toolkit for building data integration and real-time data processing pipelines.

PR #242, opened 2017-06-26
gh-216 Change dashboard title

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/216


cppwfs:
We also need to update the about.html.   I can do this on merge if you agree.

cppwfs:
LGTM

cppwfs:
Merged.

PR #246, opened 2017-07-11
[do not merge] Base stream list impl port

jvalkeal:
- This commit does a base port of most of
  a functionality.
- No flo, filtering doesn't yet work,
  paging is still partly broken, etc.
- Relates #232


sabbyanandan:
This is addressed via: #251

PR #249, opened 2017-07-12
Fix production build errors

ghillert:


ilayaperumalg:
LGTM, the build ran fine as well. Merging.

PR #251, opened 2017-07-12
Base stream list impl port plus features enumerated below

cppwfs:
Built on Janne's PR.  Here are changes:

* The properties text box will show a warning sign and change color if the properties are invalid.  (as the user types them)
* Added Toasty message for invalid deployment requests (for error messages returned from the server) .  
* Updated the handleError function to return just the message for the toasty message instead of the json error
* Removed filters for the undeploy and deploy subscribed `data` closures.  Only needed for delete
* Updated the definition page so Undeploy will be marked disabled if stream is not deploying nor deployed.

### DIFF
--- a/ui/proxy.conf.json
+++ b/ui/proxy.conf.json
@@ -9,6 +9,11 @@
         "secure": false,
         "logLevel": "debug"
     },
+    "/streams" : {
+        "target": "http://localhost:9393/",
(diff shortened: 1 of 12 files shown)


ghillert:
Merged with commit a7538dde29ebaa1520acddc3d6159b3af0675b30

PR #256, opened 2017-07-14
Fix deploying button visibility

### LINKED ISSUE #255: Disable deploy if deploying already
Deploy button should be disabled if stream is already getting deployed.

jvalkeal:
- Disabling button if stream is currently deployed
  or deploying.
- From js methods remove second index parameters
  which is undefined in template.
- Fixes #255


ilayaperumalg:
LGTM, merging.

PR #264, opened 2017-07-14
Remove unused websocket code

### LINKED ISSUE #258: About page starts non-working requests
Requests to `http://localhost:4200/websocket/info` start when visiting about page which result `404`. There requests doesn't stop even if navigating away from about page.

ui-about-bug

jvalkeal:
- Clean AboutComponent and remove websocket/stomp
  stuff which resulted runtime errors
- Fixes #258

### DIFF
--- a/ui/src/app/about/about-details.component.ts
+++ b/ui/src/app/about/about-details.component.ts
@@ -4,7 +4,6 @@ import { AboutService } from './about.service';
 import { Subscription } from 'rxjs/Subscription';
 import { BusyModule, BusyDirective } from 'angular2-busy';
 import { ToastyService, ToastyConfig, ToastOptions, ToastData} from 'ng2-toasty';
-import { StompService } from 'ng2-stomp-service';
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
@jvalkeal It looks like we also need to remove the websocket related code from `about-detail` component and the `shared` module that has a `STOMP` service.

jvalkeal:
@ilayaperumalg right ok, didn't realise there's that stuff in other places as well. I'll poke around again.

jvalkeal:
@ilayaperumalg ok, removed websocket stuff from those extra modules.

ilayaperumalg:
@jvalkeal Looks good and we need to remove the `STOMP` service and `sockjs` from `package.json`; I can do it while merging.

ilayaperumalg:
Rebased and merged as 482fd2d9e20141eba94e221ee47ca070316d0faa along with 2296836b08dbebbfb3eee109e6d81f71072d7ab6

Issue #277, opened 2017-07-24
Bulk unregister app btn still clickable even though disabled.


cppwfs:
With no apps selected on registration page the unregister button at the top shows as disabled.  However when clicking it, a modal page pops up with an empty selection list.   No action should occur in this state.

PR #278, opened 2017-07-24
gh-277 Fix Bulk unregister app btn still clickable even though disabled

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/277


cppwfs:
LGTM Squashed Merged

Issue #283, opened 2017-07-26
Move getApps Service call to Shared Module


ghillert:
As the functionality for `getApps(reload?: boolean, type?: string)` is required by issue https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/234 we should move this service call to the Share Module.

I believe feature modules should not usually depend on other feature modules.

PR #295, opened 2017-07-27
gh-283 Move getApps Service call to Shared Module

ghillert:
* Add server-side pagination support
* Add `index.ts` to `src/app/shared/model` to streamline import statements
* Improve Pagination model
  - Add `PageRequest` object
  - Add utility method `getPaginationParams` to `HttpUtils`
* Ensure that all tests pass + make the Lint gods happy

Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/283


cppwfs:
Getting the following error when clicking on register button on registration page:
```
error-handler.ts:29 Failed to convert value of type 'java.lang.String' to required type 'org.springframework.cloud.dataflow.core.ApplicationType'; nested exception is org.springframework.core.convert.ConversionFailedException: Failed to convert from type [java.lang.String] to type [@org.springframework.web.bind.annotation.PathVariable 
```

cppwfs:
I have 10 pages of apps.   When I removed apps to bring the pages from a count of 10 to 9.  Pagination bar still showed 10.

ui/src/app/shared/services/shared-apps.service.ts
```diff
@@ -0,0 +1,58 @@
...
+import { PageRequest } from '../model/pagination/page-request.model';
+import { HttpUtils } from '../support/http.utils'
+
+@Injectable()
+export class SharedAppsService {
+
+  private static appsUrl = '/apps';
+
+  constructor(private http: Http, private errorHandler: ErrorHandler) {
+    console.log('constructing');
```
cppwfs: probably don't need this log message anymore.

ui/src/app/tasks/task-apps/task-apps.component.ts
```diff
@@ -49,7 +49,7 @@ export class TaskAppsComponent implements OnInit {
   }
 
   loadAppRegistrations() {
-    this.busy = this.appsService.getApps(true, 'task').subscribe(
+    this.busy = this.tasksService.getTaskAppRegistrations().subscribe(
       data => {
         this.appRegistrations = data;
         this.toastyService.success('Task apps loaded.');
```
cppwfs: probably want to have an error closure if there is a failure.

ui/src/app/shared/components/tri-state-button.component.ts
```diff
@@ -63,6 +63,7 @@ export class TriStateButtonComponent implements AfterViewInit, DoCheck {
       this.button.nativeElement.disabled = true;
     } else {
       this.label = `Unregister all ${this._items.length} selected ${appOrApps}`;
+      this.button.nativeElement.disabled = false;
```
cppwfs: probably set the this.button.nativeElement.disabled to false on line 57 and remove the one from here and from line 60.   Will do on merge.

ui/src/app/apps/apps-register/apps-register.component.html
```diff
@@ -39,7 +39,7 @@ <h1 [ngBusy]="busy">Register Applications</h1>
                 <td [class.has-error]="type.invalid && type.touched">
                   <select #type="ngModel" name="type_{{index}}" [(ngModel)]="item.type" required>
                     <option [ngValue]="item.type" disabled="disabled">Please select a type...</option>
-                    <option *ngFor="let applicationType of applicationTypes" [ngValue]="applicationType">{{applicationType | capitalize}}</option>
+                    <option *ngFor="let applicationTypeKey of applicationType.getKeys()" [ngValue]="applicationTypeKey">{{applicationType[applicationTypeKey] | capitalize}}</option>
```
cppwfs: Source is not available in the dropdown.

ui/src/app/apps/apps.service.ts
```diff
@@ -96,7 +96,8 @@ export class AppsService {
 
     return this.http.delete(AppsService.appsUrl + '/' + appRegistration.type + '/' + appRegistration.name, options)
       .map(data => {
-        this.appRegistrations.items = this.appRegistrations.items.filter(item => item.name !== appRegistration.name);
+        const index: number = this.appRegistrations.items.findIndex(item => item.name === appRegistration.name);
+        this.appRegistrations.items.splice(index, 1);
```
cppwfs: Still seeing a slight bug here.  
While the items state is correct.  If I'm on page 9 of 10 and I delete an entry on page 9 which removes the last entry on page 10 the page count is still 10.  If we do a hard refresh then the page count goes to 9.

ui/src/app/shared/model/application-type.ts
```diff
@@ -36,12 +36,20 @@ export namespace ApplicationType {
    * ApplicationType enums.
    */
   export function getApplicationTypes() {
-    const applicationTypes: String[] = [];
+    const applicationTypes: string[] = [];
```
cppwfs: getApplicationTypes is not being used.

ui/src/app/apps/apps-register/apps-register.component.html
cppwfs: Figured this one out.  the filter in the getKeys function of ApplicationType excludes the digit 0.  My workaround was as follows:
```
   /**
   * Returns an array containing the numric key values of all
   * ApplicationType enums.
   */
  export function getKeys(): number[] {
    return Object.keys(ApplicationType).filter(isNumber).map(value => Number(value));
  }

  function isNumber(element, index, array) {
    return !isNaN(element);
  }
```

If you agree I can update on merge.

ui/src/app/shared/model/application-type.ts
cppwfs: I can remove this on merge.

ui/src/app/apps/apps.service.ts
ghillert: @cppwfs Pleas create an issue for that one -- I need to think about that one a bit. When on the same page and you delete things, I am not doing page refreshes for single item deletes, I think for absolutely perfect and consistent behavior that needs to change, and we need to do a server-call to refresh the page.

ghillert:
@cppwfs Rebased + polished

cppwfs:
LGTM.   Squashed Merged

PR #309, opened 2017-08-01
Remove sample text

### LINKED ISSUE #307: Delete stream-dialog includes confusing text
As a user, I tried destroying a stream and I see the following text in the dialog window, which is confusing. My stream definition was named as `foo`.

- What is `foostream1`? 
- Why are we highlighting `foo` with an underline? (it made me think I can click it, but it is not an hyperlink)

image

cppwfs:
resolves #307


ghillert:
LGTM - merging.

ghillert:
Merged with commit 5d722bbbd5cdb9288f33aafcd50a15b5de24a05f

Issue #311, opened 2017-08-01
Fix Production Bundle Error


ghillert:

ghillert:
Issued is related to https://github.com/devyumao/angular2-busy/issues/55

The temporary fix is to use https://www.npmjs.com/package/tixif-ngx-busy for the time being. 

In the course of the investigation and resolution, I upgraded to Angular CLI `1.3.0 RC3`.  That should make the upgrade to the final release just a version flip when addressing:

https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/285

Issue #285, opened 2017-07-26
Update to Angular CLI version to 1.3.0


cppwfs:
Also update the Angular 4.0.0 in package.json to 4.3.0

ghillert:
See also: https://github.com/angular/angular-cli/releases

ghillert:
We had already upgraded previously to the 1.3.0 RC release of Angular CLI. So this is just a version bump.

PR #312, opened 2017-08-02
gh-285 Fix Production bundle error

ghillert:
- Temporarily upgrade to `tixif-ngx-busy`
- Upgrade to Angular Cli `1.3.0 RC3`

Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/311


cppwfs:
Once this PR is merged.  We will need to put a note in the slack room telling folks to 
1) delete node_modules 
2) npm install
3) ng build --prod

ui/package.json
```diff
@@ -11,39 +11,42 @@
...
+    "@angular/forms": "^4.2.4",
+    "@angular/http": "^4.2.4",
+    "@angular/platform-browser": "^4.2.4",
+    "@angular/platform-browser-dynamic": "^4.2.4",
+    "@angular/router": "^4.2.4",
     "core-js": "^2.4.1",
+    "rxjs": "^5.4.2",
+    "zone.js": "^0.8.14",
+
+    "tixif-ngx-busy": "0.0.5",
```
cppwfs: Why did we switch from angular2-busy which is at 2.0.4 to tixif-ngx-busy with is at 0.0.5 ?

cppwfs:
As a separate issue we probably want to update our build instructions on the readme.  it says to `mvn clean package install`  Which causes 2 compodoc builds. probably want switch to `npm clean install`

ui/src/polyfills.ts
```diff
@@ -32,21 +31,21 @@ import 'core-js/es6/string';
 import 'core-js/es6/array';
 // import 'core-js/es6/regexp';
 // import 'core-js/es6/map';
+// import 'core-js/es6/weak-map';
```
ghillert: They are part of the Angular CLI files. Thus I would rather leave the Angular provided files as unmodified as possible.  That way we know also all the relevant polyfills and can enabled them as needed we start testing across more browsers.

ui/tslint.json
```diff
@@ -80,6 +86,7 @@
     ],
     "radix": true,
     "semicolon": [
+      true,
```
ghillert: Sounds good - please create a follow up issue as it would affect a large number of files.

ui/package.json
ghillert: Issued is related to https://github.com/devyumao/angular2-busy/issues/55

The temporary fix is to use https://www.npmjs.com/package/tixif-ngx-busy for the time being.

cppwfs:
LGTM.  Rebased Merged.

Issue #298, opened 2017-07-31
Page Count incorrect on Apps tab after delete


cppwfs:
If I'm on page 9 of 10 and I delete an entry on page 9 which removes the last entry on page 10 the page count is still 10. If we do a hard refresh then the page count goes to 9.

PR #316, opened 2017-08-02
gh-298 Fix Page Count incorrect on Apps tab after delete

ghillert:
* Provide doc comments for all methods of the `AppsComponent`

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/298

### DIFF
--- a/ui/src/app/apps/apps.component.ts
+++ b/ui/src/app/apps/apps.component.ts
@@ -24,7 +24,7 @@ import { ModalDirective } from 'ngx-bootstrap/modal';
 export class AppsComponent implements OnInit {
 
   appRegistrations: Page<AppRegistration>;
-  busy: Subscription[] = [];
+  busy: Subscription;
 
   appRegistrationToUnregister: AppRegistration;
   appRegistrationsToUnregister: AppRegistration[];
@@ -56,7 +56,7 @@ export class AppsComponent implements OnInit {
    * @param reload
    */
   public loadAppRegistrations(reload: boolean) {
-    this.busy.push(this.appsService.getApps(reload).subscribe(
+    this.busy = this.appsService.getApps(reload).subscribe(
       data => {
         if (!this.appRegistrations) {
           this.appRegistrations = data;
@@ -66,7 +66,7 @@ export class AppsComponent implements OnInit {
         console.log('error', error);
         this.toastyService.error(error);
       }
-    ));
+    );
   }
 
   /**
@@ -119,7 +119,7 @@ export class AppsComponent implements OnInit {
   public proceedToUnregisterSingleAppRegistration(appRegistration: AppRegistration): void {
     console.log('Proceeding to unregister application...', appRegistration);
 
-    this.busy.push(this.appsService.unregisterApp(appRegistration).subscribe(
+    this.busy = this.appsService.unregisterApp(appRegistration).subscribe(
       data => {
         this.unregisterSingleAppModal.hide();
         this.toastyService.success('Successfully removed app "'
@@ -128,15 +128,15 @@ export class AppsComponent implements OnInit {
         if (this.appsService.appRegistrations.items.length === 0 && this.appsService.appRegistrations.pageNumber > 0) {
           this.appRegistrations.pageNumber = this.appRegistrations.pageNumber - 1;
         }
-        const reloadAppsObservable = this.appsService.getApps(true).subscribe(
+        const reloadAppsSubscription = this.appsService.getApps(true).subscribe(
           appRegistrations => {}
         );
-        this.busy.push(reloadAppsObservable);
+        this.busy = reloadAppsSubscription;
       },
       error => {
         this.toastyService.error(error);
       }
-    ));
+    );
   }
 
   /**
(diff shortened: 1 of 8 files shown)


ui/src/app/apps/apps.component.ts
```diff
@@ -35,12 +42,21 @@ export class AppsComponent implements OnInit {
...
   }
 
+  /**
+   * Load a paginated list of {@link AppRegistration}s.
+   *
+   * @param reload
+   */
   public loadAppRegistrations(reload: boolean) {
-    this.busy = this.appsService.getApps(reload).subscribe(
+    this.busy.push(this.appsService.getApps(reload).subscribe(
```
jvalkeal: Is it going to be a problem that we push to array but never clear it?

jvalkeal: While you stay on apps page that array just keeps growing when you go on different pages, while moving between tabs will re-init a class. I'd prefer not to leave this type of code here as possible copy/pasting will then introduce it to somewhere else. At minimum we should create an issue to track this if we merge without a fix.

ui/src/app/apps/apps.service.ts
```diff
@@ -102,6 +102,14 @@ export class AppsService {
 
   }
 
+  unregisterMultipleApps(appRegs: AppRegistration[]): Observable<Response[]> {
+    const observables: Observable<Response>[] = [];
+    for (const appReg of appRegs) {
+      observables.push(this.unregisterApp(appReg));
+    }
+    return Observable.forkJoin(observables);
```
jvalkeal: While using `forkJoin` is not a new thing in this class, doesn't this have a problem that if one `Observable` in forkjoin fails, then whole `Observable` errors and rest in forkjoin are not even tried. Probably not an issue for this PR but we should think about how this behaves when things go south.

ui/src/app/apps/apps.component.ts
ghillert: @jvalkeal It turns out, this was not working correctly anyway. I assumed the busy indicator would work smarter then it does in actuality. Ideally, my page may at random create Observables whose Subscriptions I would like to add to the busy indicator. Therefore, "overlapping Subscriptions" could be a possibility.

Thus, it would have been nice if Busy Indicator could consume a stateful array and upon completion remove Subscriptions from the array. 

Maybe I am off here but just to double-check, I created an issue in the Angular2 Busy issue tracker - https://github.com/devyumao/angular2-busy/issues/77

ui/src/app/apps/apps.service.ts
ghillert: @jvalkeal Yes we need to polish those things as we go along and add catch-statements in scenarios where we need to have some "continue on error" logic.

E.g. see here: https://stackoverflow.com/a/38061516/835934

In this particular use-case, I think, we are borderline-ok.

ui/src/app/apps/apps.component.ts
```diff
@@ -86,31 +127,67 @@ export class AppsComponent implements OnInit {
...
+    const subscription = this.appsService.unregisterMultipleApps(appRegistrations).subscribe(
+      data => {
+        console.log(data);
+        this.toastyService.success(`${data.length} app(s) unregistered.`);
+
+        if (this.appsService.appRegistrations.items.length === 0 && this.appsService.appRegistrations.pageNumber > 0) {
+          this.appRegistrations.pageNumber = this.appRegistrations.pageNumber - 1;
+        }
+        const reloadAppsSubscription = this.appsService.getApps(true).subscribe(
+          appRegistrations => {}
```
cppwfs: This is getting tagged by lint as a shadowed variable?  Just change to appReg or something else.

ui/src/app/apps/apps.component.ts
```diff
@@ -86,31 +127,67 @@ export class AppsComponent implements OnInit {
 
         if (this.appsService.appRegistrations.items.length === 0 && this.appsService.appRegistrations.pageNumber > 0) {
           this.appRegistrations.pageNumber = this.appRegistrations.pageNumber - 1;
-          this.loadAppRegistrations(true);
         }
+        const reloadAppsSubscription = this.appsService.getApps(true).subscribe(
```
cppwfs: `this.busy` can be used instead of `reloadAppsSubscription`

ui/src/app/apps/apps.component.ts
```diff
@@ -86,31 +127,67 @@ export class AppsComponent implements OnInit {
...
-    }
+    const subscription = this.appsService.unregisterMultipleApps(appRegistrations).subscribe(
+      data => {
+        console.log(data);
+        this.toastyService.success(`${data.length} app(s) unregistered.`);
+
+        if (this.appsService.appRegistrations.items.length === 0 && this.appsService.appRegistrations.pageNumber > 0) {
+          this.appRegistrations.pageNumber = this.appRegistrations.pageNumber - 1;
+        }
+        const reloadAppsSubscription = this.appsService.getApps(true).subscribe(
```
cppwfs: this.busy can be used instead of reloadAppsSubscription.  This can be applied to the others.

ghillert:
@jvalkeal @cppwfs Addressed comments.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/13f72ec35b180222d01a4a46347744bba098dd76

Issue #313, opened 2017-08-02
Resolve ng lint failures


cppwfs:
Do to changes in lint config in issue #311, the ng lint fails.

PR #318, opened 2017-08-02
[REVIEW ONLY] gh-313 Resolve ng lint failures

ghillert:
@cppwfs A couple of errors remain. Ran out of time tonight. Maybe you can pick up from here. Or create separate issues.

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/313

### DIFF
--- a/ui/src/app/about/about-details.component.ts
+++ b/ui/src/app/about/about-details.component.ts
@@ -41,7 +41,7 @@ export class AboutDetailsComponent implements OnInit {
   goBack() {
     console.log('Back to about page ...');
     this.router.navigate(['about']);
-  };
+  }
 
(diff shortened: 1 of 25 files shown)


cppwfs:
Added one more fix to a for loop in a pipe.  When using a map in a for loop (to get a list of keys) you need to wrap the map in a Object.keys function.

cppwfs:
With your changes and my addition, that resolves all the code style and some bugs, the remaining are actually bugs that lint identified and I'll enumerate those in separate stories.

cppwfs:
LGTM

cppwfs:
Added some fixes.  Squashed, rebased, merged.

Issue #338, opened 2017-08-07
Add test coverage for Jobs


sabbyanandan:
As a developer, I'd like to improve the test coverage for jobs tab.

ghillert:
See also https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/345

PR #345, opened 2017-08-10
gh-238 Refactor and port jobs list view to Angular 4

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/238

### DIFF
--- a/ui/src/app/jobs/jobs.service.ts
+++ b/ui/src/app/jobs/jobs.service.ts
@@ -1,39 +1,83 @@
 import { Injectable } from '@angular/core';
-import { Http, Response } from '@angular/http';
+import { Http, Response, RequestOptionsArgs } from '@angular/http';
+
 import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/operator/catch';
 import 'rxjs/add/operator/map';
 
+import { ErrorHandler, Page } from '../shared/model';
+import { PageRequest } from '../shared/model/pagination/page-request.model';
+
+import { JobExecution } from './model/job-execution.model';
+import { HttpUtils } from '../shared/support/http.utils';
+
+import * as moment from 'moment';
+import { Moment } from 'moment';
+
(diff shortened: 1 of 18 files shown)


jvalkeal:
Overall looks relatively ok. I believe there's some preliminary work to prepare future changes when (possibly step executions or composed tasks are added). I'd prefer to merge this sooner rather than later as I also need that moment.js stuff. But was there a plan to add better tests or shall we handle those in a separate issue?

ui/src/app/jobs/jobs.service.ts
```diff
@@ -1,39 +1,83 @@
...
-    // In a real world app, you might use a remote logging infrastructure
-    let errMsg: string;
-    if (error instanceof Response) {
-      const body = error.json() || '';
-      const err = body.error || JSON.stringify(body);
-      errMsg = `${error.status} - ${error.statusText || ''} ${err}`;
     } else {
-      errMsg = error.message ? error.message : error.toString();
+      this.remotelyLoaded = false;
+      console.log('Fetching Job Executions from local state.', this.jobExecutions);
```
cppwfs: With Job Executions we want to see the latest state upon tab selection, not the saved state.  Because the job execution state could have changed since the last time they were on the tab.

ghillert: made this change 

```
ngOnInit() {
    this.loadJobExecutions(true);
}
```

ghillert:
Addressed most code review changes + rebased. Will add more tests.

ghillert:
@jvalkeal Will add more tests as part of https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/338

A good reminder to create smaller feature stories.

cppwfs:
LGTM

cppwfs:
Rebased, Lint Polishing, Squashed, Merged.

PR #361, opened 2017-08-17
Adds tests for invalid Step and Job Execution ID for detail page.

### LINKED ISSUE #360: Add tests to handle invalid jobId and Step Id for JobExecutionDetail
And StepExecution Detail

cppwfs:
* Also fixes the templates so tables and headers that should not appear on empty pages will not show.
* Also Adds "No Step Execution" message for the Step Execution Page.

resolves #360


jvalkeal:
lgtm, merging

PR #370, opened 2017-08-23
Add toasty message for duplicate definition

### LINKED ISSUE #341: Creating task definition with duplicate name shows no Toasty Error
When creating a task definition using an existing task definition name, no error message is presented.

cppwfs:
if definition already exists the server will return an error message and this PR will display the toasty message.

resolves #341


ghillert:
Merged with commit 9baad3d66e44ce36f01a2259160d2b9b87d6bc71

PR #376, opened 2017-08-29
Add sorting to task definitions tables

### LINKED ISSUE #275: Add sorting to task definitions tables
New angular UI doesn't yet support sorting as old one do.

jvalkeal:
- Add style font-awesome.min.css to get sorting arrows.
- Define individual states for sorting by name
  and definition. Toggle between null, true and false
  to indicate no sort, desc and asc.
- Modify tasks service to pass sorting param to where
  sorting by definition is primary as sort by name would
  be pointless as names are unique. Columns sort order has
  to be fixed as in UI it'd be too awkward to define whether
  name or definition is use as a primary sort.
- This commit only touches definitions table(not task apps table)
  to see if concept is ok to use in all other tables.
- Fixes #275

### DIFF
--- a/ui/src/app/tasks/task-definitions/task-definitions.component.ts
+++ b/ui/src/app/tasks/task-definitions/task-definitions.component.ts
@@ -18,6 +18,13 @@ export class TaskDefinitionsComponent implements OnInit {
   busy: Subscription;
   taskDefinitionToDestroy: TaskDefinition;
 
+  // null indicates that order is not selected
+  // thus we simply show double arrow.
+  // for sorting toggles, we just switch
+  // and first sort with true which equals desc
+  definitionNameSort: boolean = null;
+  definitionSort: boolean = null;
+
   @ViewChild('childPopover')
   public childPopover: PopoverDirective;
 
@@ -50,7 +57,7 @@ export class TaskDefinitionsComponent implements OnInit {
   loadTaskDefinitions() {
     console.log('Loading Task Definitions...', this.taskDefinitions);
 
-    this.busy = this.tasksService.getDefinitions().subscribe(
+    this.busy = this.tasksService.getDefinitions(this.definitionNameSort, this.definitionSort).subscribe(
       data => {
         this.taskDefinitions = data;
         this.toastyService.success('Task definitions loaded.');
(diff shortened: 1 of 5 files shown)


ui/src/app/tasks/tasks.service.spec.ts
```diff
@@ -37,4 +37,43 @@ describe('TasksService', () => {
     });
   });
 
+  describe('getDefinitions', () => {
+    it('should call the definitions service with the right url [no sort params]', () => {
+      this.mockHttp.get.and.returnValue(Observable.of(this.jsonData));
+
+      const params: URLSearchParams = HttpUtils.getPaginationParams(0, 10);
+      // params.append('type', 'task');
```
cppwfs: remove this commented line.

cppwfs:
Overall LGTM.  But let's  @sabbyanandan @ghillert @BoykoAlex take a peak and give feedback.

ui/src/app/tasks/task-definitions/task-definitions.component.ts
```diff
@@ -18,6 +18,13 @@ export class TaskDefinitionsComponent implements OnInit {
   busy: Subscription;
   taskDefinitionToDestroy: TaskDefinition;
 
+  // null indicates that order is not selected
+  // thus we simply show double arrow.
+  // for sorting toggles, we just switch
+  // and first sort with true which equals desc
+  definitionNameSort: boolean = null;
```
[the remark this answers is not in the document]
ghillert: See also https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

ui/src/app/tasks/task-definitions/task-definitions.component.ts
```diff
@@ -18,6 +18,13 @@ export class TaskDefinitionsComponent implements OnInit {
...
+  definitionSort: boolean = null;
```
ghillert: I wonder if we should make that somehow part of Pageable - see Spring Data's `org.springframework.data.domain.PageRequest` and also have it be part of the service in order to have some stateful behavior.

cppwfs:
Polished, Rebased, Squashed, Merged.

PR #382, opened 2017-08-31
Change table layout css

### LINKED ISSUE #359: Step Name column on StepExecutions list on Job Execution Detail page spills over into other column
From the Job's tab click a magnifying glass button by one of the job executions.  Scroll to the bottom of the page.  When viewing the `Step Name` in the Steps table, if the step name is like `workerStep:partition2` it will spill over into the `Reads` column.

jvalkeal:
- Changing table-layout from fixed back to default
  auto which is set by PUI. Fixed layout just doesn't
  seem to work overall if seeing all tables we use,
  ecpecially when narrowing browser size.
- While fixed might be better in some cases, we should
  not use it until we know better settings which
  works with all tables.
- Fixes #359

### DIFF
--- a/ui/src/styles.scss
+++ b/ui/src/styles.scss
@@ -55,6 +55,14 @@ $icon-font-path: '~bootstrap-sass/assets/fonts/bootstrap/';
 
 @import '../node_modules/@compodoc/compodoc/src/resources/styles/font-awesome.min.css';
 
+// PUI sets this to fixed which completely
+// breaks table behaviour. With default setting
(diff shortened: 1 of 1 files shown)


cppwfs:
LGTM

cppwfs:
Did find one issue (had to pull the LGTM :-( ), the stream definitions page buttons layout needs to be changed with the table layout set to auto.   Do we want to resolve this issue with this PR or open another story?

cppwfs:
re-adding LGTM :-)

PR #387, opened 2017-08-31
Add sorting to stream definitions table

### LINKED ISSUE #385: Add sorting to stream definitions table

jvalkeal:
- Keep sorting states for definition and name
  which are added to request in this order.
- Fixes #385

### DIFF
--- a/ui/src/app/streams/streams.service.ts
+++ b/ui/src/app/streams/streams.service.ts
@@ -37,10 +37,13 @@ export class StreamsService {
 
   /**
    * Retrieves the {@link StreamDefinition}s based on the page requested.
+   *
+   * @param definitionNameSort the sort for DEFINITION_NAME
+   * @param definitionSort the sort for DEFINITION
    * @returns {Observable<R|T>} that will call the subscribed funtions to handle
    * the results when returned from the Spring Cloud Data Flow server.
    */
-  getDefinitions(): Observable<Page<StreamDefinition>> {
+  getDefinitions(definitionNameSort?: boolean, definitionSort?: boolean): Observable<Page<StreamDefinition>> {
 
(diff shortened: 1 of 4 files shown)


ui/src/app/streams/streams.service.ts
```diff
@@ -49,8 +52,21 @@ export class StreamsService {
       this.streamDefinitions.pageNumber,
       this.streamDefinitions.pageSize
     );
-      // TODO Implement Sorting
-      // params.sort = pageable.calculateSortParameter();
+
+    if (definitionSort != null) {
```
cppwfs: undefined instead of null.  Will fix on merge.

ui/src/app/streams/streams.service.ts
```diff
@@ -49,8 +52,21 @@ export class StreamsService {
...
+      if (definitionSort) {
+        params.append('sort', 'DEFINITION,DESC');
+      } else {
+        params.append('sort', 'DEFINITION,ASC');
+      }
+    }
+    if (definitionNameSort != null) {
```
cppwfs: undefined instead of null.  Will fix on merge.

ui/src/app/streams/streams.service.ts
jvalkeal: Ah, we didn't change these in TasksService and as model was same I just copy/pasted. Weird as I remember some nulls were changed to undefined with that older PR.

cppwfs: Updating the task service as a part of the polishing on this PR.

cppwfs:
LGTM

cppwfs:
Replaced `null`s in Stream and task services with `undefined`s, polished, rebased, squashed, merged.

PR #388, opened 2017-08-31
Flo for Streams creation (M2)

BoykoAlex:
Flo editor for stream creation in a usable form. 
Main limitations:
* No squigglies in the DSL editor
* No tooltips
* No code snippet property form editor control

Work in progress. This delivery should allow for `Bulk Task Define` feature to be ported to Anfular 4 and TS
Test coverage should be improved as closer we get to RC

Would be nice to get feedback on directory/file layout first.


jvalkeal:
Just first generic comments before I look this any further

1. Lot of commented code which should be removed.

2. `ng lint` gives 471 ERRORS

3. graph-to-text destination source test fails:
```
Expected ':d1 > log
:d1 > log
:d1 > file' to equal ':d1 > log
:d1 > file'.
```

BoykoAlex:
@jvalkeal the test fails because it's work in progress. It catches something that is not working

jvalkeal:
What comes for naming ts files, we should probably follow model i.e. where `StreamCreateDialogComponent` is in file `stream-create-dialog.component.ts`, not `stream.create.dialog.component.ts`.

What comes for lint errors, I believe those will fail actual build.

BoykoAlex:
@jvalkeal Hmm... I think i saw maven build succeeding with all these linting problems... Linting is a bit frustrating... for example indents are expected to be spaces rather than tabs... also `a : boolean;` is expected to be `a: boolean;`...
I commented out the failing work-in-progress test, lets see if build passes.

jvalkeal:
hmm, I thought maven would fail with lint. Anyway we're using lint to catch some potential issues and it keeps code in some sane formatting so that next guy can create more pretty PR. It's annoying but if we let it slip then whole lint system becomes pointless.

sabbyanandan:
@ghillert @jvalkeal Why does the build pass when there is `ng lint` errors?

jvalkeal:
@sabbyanandan it might be that @cppwfs has acted as a gatekeeper and not letting any PR's to go through if lint errors. That why I thought a real build would fail. Anyway @ghillert do you have any immediate comments for this PR before I start thinking about merging it? Point is to get into safe point to merge it and then handle remaining issues later.

cppwfs:
Yeah, it was me ;-).  I either fixed them on commit or added nags on the PR.

jvalkeal:
I'm squashing, doing a bit polishing and merging. Lets not over-engineer this PR, instead fix some issues in future commits as this was anyway work-in-progress.

jvalkeal:
Closing. Squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/381365b860ac39a9868636261011508bc3d78918. Also quick polish per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/8847131b18e4c90cd4811cc50763c7d56f146376.

PR #389, opened 2017-09-02
[POC DO NOT MERGE] ngrx/store for tasks definitions page

jvalkeal:
- You need to `npm update` due changes in package.json.
- Play with filtering, paging and sorting, go back to other tab
  and come back to definitions tab. State should be restored.
- Some good resources:
  https://github.com/ngrx/platform/blob/master/docs/store/README.md
  https://github.com/ngrx/platform/tree/master/example-app
  https://www.sitepoint.com/managing-state-angular-2-ngrx
- Showing concept of using ngrx store to keep global hierachical
  states with impl of state for tasks definitions page.
- State for filter, sorting and pagination are kept in store
  so that when you nagivate out from a page and go back,
  state is restored.
- Pagination state removed from TasksService.
- Adding ngrx store, recusers, actions and effects.
- Whole state for tasks definitions are kept in store
  and asynchronously selected from there and updated
  into a page structure.
- There are different action to send raw inputs for filter,
  bounce that filter action, sorts, paging, update and reset.
- As a simple POC, doesn't address tests!

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -24,6 +24,8 @@
     "@angular/platform-browser": "^4.2.4",
     "@angular/platform-browser-dynamic": "^4.2.4",
     "@angular/router": "^4.2.4",
+    "@ngrx/store": "^4.0.0",
+    "@ngrx/effects": "^4.0.0",
     "core-js": "^2.4.1",
     "rxjs": "^5.4.2",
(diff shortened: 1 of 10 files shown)


sabbyanandan:
Since we haven't revisited this topic for a while, I will close this for now.

PR #407, opened 2017-09-05
Switch to specific commit of Flo. Put back removed commented out code

BoykoAlex:
1. Switch to specific commit of Spring-Flo instead of master branch. I'll allow us to work on Flo and SCDF-UI at the same time without breaking SCDF-UI master code
2. Put back some of removed commented out code. Mostly it's for the code that just lists some example JSON such that it's easy to understand what is being tested.

### DIFF
--- a/ui/src/app/shared/services/parser.spec.ts
+++ b/ui/src/app/shared/services/parser.spec.ts
@@ -14,6 +14,9 @@ describe('parser:', () => {
     line = parseResult.lines[0];
     expect(line.errors).toBeNull();
     node = line.nodes[0];
+    // [{"group":"UNKNOWN_0","type":"sink","name":"time","options":{},"optionsranges":{},
+    //   "range":{"start":{"ch":0,"line":0},"end":{"ch":4,"line":0}},
+    //   "sourceChannelName":null,"sinkChannelName":null}]
     expect(node.group).toEqual('UNKNOWN_0');
     expect(node.type).toEqual('source');
     expect(node.name).toEqual('time');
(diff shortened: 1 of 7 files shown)


ui/src/app/shared/services/parser.spec.ts
```diff
@@ -14,6 +14,9 @@ describe('parser:', () => {
     line = parseResult.lines[0];
     expect(line.errors).toBeNull();
     node = line.nodes[0];
+    // [{"group":"UNKNOWN_0","type":"sink","name":"time","options":{},"optionsranges":{},
```
cppwfs: May want to JSDoc these so that you can explain the JSON.

ui/src/app/shared/services/tokenizer.ts
```diff
@@ -296,8 +306,8 @@ class Tokenizer {
                     this.lexDoubleQuotedStringLiteral();
                     break;
                 // case '@':
-                // this.pushCharToken(TokenKind.REFERENCE);
-                // break;
+                // 	this.pushCharToken(TokenKind.REFERENCE);
```
jvalkeal: Lets put these back if/when those are actually used.

ui/src/app/shared/services/parser.spec.ts
jvalkeal: Might be better to create a full expected mock object somewhere else and then assert against that. I really don't see this very useful to put some long examples in comments and then assert some fields from it. Also makes code to look not that pretty.

ui/src/app/shared/services/tokenizer.ts
BoykoAlex: Agreed. Removed it.

ui/src/app/shared/services/parser.spec.ts
BoykoAlex: Comments in tests reworked into asserts.

BoykoAlex:
@cppwfs @jvalkeal I have put in adjustments to the PR.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/ea9af0a7f75580bd4042c3b7bb35b642b355f5a1

Issue #322, opened 2017-08-02
Refactor and port Analytics Counter tab


sabbyanandan:
As a developer, I'd like to refactor and port the Analytics tab to the new Angular4 infrastructure.

PR #410, opened 2017-09-06
gh-322 Refactor and port Analytics Counter tab

ghillert:
* Create Analytics component
* Create Counter component
* Add routing configuration
* Migrate Counter Line Chart component to Typescript and D3v4

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/322


ui/src/app/analytics/analytics.component.spec.ts
```diff
@@ -19,7 +19,7 @@ xdescribe('AnalyticsComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should be created', () => {
+  xit('should be created', () => {
```
jvalkeal: What's a reason to disable test?

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
 
-  constructor(private http: Http) { }
+  constructor(private http: Http, private errorHandler: ErrorHandler) {
+  }
 
-  getApps(): Observable<any[]> {
-    return this.http.get(this.appstUrl)
-                    .map(this.extractData)
-                    .catch(this.handleError);
+  set counterIntervall(rate: number) {
```
jvalkeal: typo, extra `l` at the end.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
+          this.startPollingForCounters();
+        }
+      }
+    }
   }
 
-  private extractData(res: Response) {
-    const body = res.json();
-    return body._embedded;
+  get counterIntervall(): number {
```
jvalkeal: typo, extra `l` at the end.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
-    } else {
-      errMsg = error.message ? error.message : error.toString();
+  public totalCacheSize() {
+    return Math.max(Math.ceil(60 / this._counterInterval), 20);
+  }
+
+  startPollingForCounters() {
+    if (!this.counterPoller || this.counterPoller.closed) {
+      this.counterPoller = Observable.interval(this._counterInterval * 1000)
+        .switchMap(() => this.getAllCounters(true)).subscribe();
```
jvalkeal: If you don't have redis setup, backend service errors
```
$ curl "http://localhost:9393/metrics/counters?page=0&size=10&detailed=true"
[{"logref":"RedisConnectionFailureException","message":"Cannot get Jedis connection; nested exception is redis.clients.jedis.exceptions.JedisConnectionException: Could not get a resource from the pool","links":[]}]
```
I think this is causing whole Observable to fail and no further http requests are made.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
-      errMsg = `${error.status} - ${error.statusText || ''} ${err}`;
-    } else {
-      errMsg = error.message ? error.message : error.toString();
+  public totalCacheSize() {
+    return Math.max(Math.ceil(60 / this._counterInterval), 20);
+  }
+
+  startPollingForCounters() {
+    if (!this.counterPoller || this.counterPoller.closed) {
+      this.counterPoller = Observable.interval(this._counterInterval * 1000)
```
jvalkeal: We can't just use interval as is because if you set it to 1 sec and actual request take more that 1 sec, you have instant memory leak as things will queue up. Not sure if there's any other operator in rxjs(as we have in reactor) to schedule when downstream have done its stuff. Anyway logic should be that interval is between end of succeed/failed request and start of next request.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
-    // In a real world app, you might use a remote logging infrastructure
-    let errMsg: string;
-    if (error instanceof Response) {
-      const body = error.json() || '';
-      const err = body.error || JSON.stringify(body);
-      errMsg = `${error.status} - ${error.statusText || ''} ${err}`;
-    } else {
-      errMsg = error.message ? error.message : error.toString();
+  public totalCacheSize() {
+    return Math.max(Math.ceil(60 / this._counterInterval), 20);
```
jvalkeal: Maybe just a nitpicking but if we have a property getter/setter, then access to its value should be via getter/setter and not short-circuit bypassing those.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
+  startPollingForCounters() {
+    if (!this.counterPoller || this.counterPoller.closed) {
+      this.counterPoller = Observable.interval(this._counterInterval * 1000)
+        .switchMap(() => this.getAllCounters(true)).subscribe();
     }
-    console.error(errMsg);
-    return Observable.throw(errMsg);
   }
 
+  stopPollingForCounters() {
```
jvalkeal: I don't see this called from anywhere if you navigate away from analytics. Thus unneeded requests are made unless you actually go back to counters tab and set interval to 0. Setting interval to 0 stop poller but then it's not reactivated if you set interval to 1(thus you need to go some other tab and go back to counters tab which activates poller again)

ui/src/app/analytics/analytics.component.html
```diff
@@ -1,4 +1,17 @@
...
+
+<div class="tab-simple">
+  <ul class="nav nav-tabs">
+    <li role="presentation" routerLinkActive="active"><a routerLink="dashboard">Dashboard</a></li>
+    <li role="presentation" routerLinkActive="active"><a routerLink="counters">Counters</a></li>
+  </ul>
+  <div class="tab-content">
+    <div class="tab-pane in active"><router-outlet></router-outlet></div>
+  </div>
+</div>
```
cppwfs: carriage return.

ui/src/app/analytics/analytics.component.spec.ts
cppwfs: @ghillert update your `beforeEach` with this code segment and the test will work:
```
  beforeEach(async(() => {
    activeRoute = new MockActivatedRoute();
    TestBed.configureTestingModule({
      declarations: [ AnalyticsComponent ],
      imports: [
      RouterTestingModule.withRoutes([])
      ],
      providers: [
        { provide: ActivatedRoute, useValue: activeRoute }
      ]
    })
    .compileComponents();
  }));
```

cppwfs: @ghillert   update your `beforeEach` with the segment below and the test will work.
```
beforeEach(async(() => {
    activeRoute = new MockActivatedRoute();
    TestBed.configureTestingModule({
      declarations: [ AnalyticsComponent ],
      imports: [
      RouterTestingModule.withRoutes([])
      ],
      providers: [
        { provide: ActivatedRoute, useValue: activeRoute }
      ]
    })
    .compileComponents();
  }));
```

ui/src/app/analytics/analytics.service.ts
cppwfs: @jvalkeal Can you clarify the test case.  I killed my instance of redis,  refreshed the page and the counters disappeared.  Once I started the redis instance , and refreshed the page the counters were back..

cppwfs: We also need to add the following to the subscribe()
```
    error => {
        this.toastyService.error(error);
      }
```

ui/src/app/analytics/counters/counters.component.ts
```diff
@@ -0,0 +1,75 @@
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { Router } from '@angular/router';
+
+import { Subscription } from 'rxjs/Subscription';
```
cppwfs: Can remove subscription, ViewChild, Page, ModalDirective imports.  Unused.

ui/src/app/analytics/counters/counters.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  public refreshRate: number;
+
+  get counters() {
+    return this.analyticsService.counters;
+  }
+
+  constructor(
+    public analyticsService: AnalyticsService,
+    private toastyService: ToastyService,
```
cppwfs: Toasty Service and Router are not used.

ui/src/app/analytics/counters/counters.component.ts
```diff
@@ -0,0 +1,75 @@
...
+    console.log(`Getting page ${page}.`);
+    this.analyticsService.counters.pageNumber = page - 1;
+  }
+
+  public totalCacheSize(): number {
+    return this.analyticsService.totalCacheSize();
+  }
+
+  public changeRefreshRate() {
+    console.log('Changing refreshrate to ...' + this.refreshRate);
```
cppwfs: change `refreshrate` to refreshRate

cppwfs:
One annoying thing that I found.
I have 3 counters.  A tweets counter, a counter hooked to a time source and counter that is idle.
The page keeps scrolling to the bottom.  But doesn't keep pace with the 2 second time refresh.   It seems a bit slower.

ui/src/app/analytics/analytics.service.ts
```diff
@@ -1,39 +1,115 @@
...
-    return this.http.get(this.appstUrl)
-                    .map(this.extractData)
-                    .catch(this.handleError);
+  set counterIntervall(rate: number) {
+    if (rate && !isNaN(rate)) {
+      if (rate < 0.01) {
+          rate = 0;
+          this.stopPollingForCounters();
+      } else {
+        console.log('Setting intervall to ' + rate);
```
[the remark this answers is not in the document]
jvalkeal: maybe that's some sort of Hawaii thing ;)

cppwfs:
When setting update seconds to zero.  All the x access markers are set to 0 as well.   We should maybe leave the previous intervals.

cppwfs:
After setting interval to 0.   setting it back to 2 seconds does not reactivate the graph.  It stays paused.

jvalkeal:
I see that page move as well. I may be cause by how that table is updated as I assume if chart is re-rendered, its content is removed and added. That might cause table size to change thus causing whole page size to change. At some point I also saw table row flickering when it was updated.

ui/src/app/analytics/analytics.service.ts
ghillert: Created a backend story for this also: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1636

ui/src/app/analytics/analytics.component.html
ghillert: Fixed.

ui/src/app/analytics/charts/graph-chart/graph-chart.component.html
```diff
@@ -0,0 +1 @@
+<div class="d3-chart" #chart></div>
```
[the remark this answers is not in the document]
ghillert: Fixed

ghillert:
Regarding the poller - I created https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/417

jvalkeal:
ok, I'm going to merge with some polish and will create some additional issues.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/8c87b2252a3f4770477551b19c955ad861df76cc

PR #411, opened 2017-09-06
Switch action buttons to icons

### LINKED ISSUE #384: Switch action buttons to icons
After setting the table layout to auto, the deploy, destroy, undeploy, & detail buttons on the list pages are misaligned.   After a discussion between Sabby, Glenn & Janne, it was decided that they should be icons instead of buttons with text.  With hover over text the meaning of each button should be enough to allow the user to know what the purpose of the button is.

This model needs to be applied in following pages:

- Apps
- Stream List
- Task List
- Job List
- Execution List

jvalkeal:
- Remove labels and just use icons.
- Polish table styles as things look
  a bit weird when window is narrow.
- Fixes #384


cppwfs:
LGTM

cppwfs:
Merged

PR #421, opened 2017-09-07
Encoder of HTTP request parameters

### LINKED ISSUE #413: TAP'ed streams don't work when created/deployed from Flo
As a user, I deployed the following streams from Flo and it doesn't appear to be working as expected. The TAP'ed streams specifically don't appear to be connected to the same consumer-group, so there are no new messages in `STREAM_2` and `STREAM_3` log-apps. 

```
STREAM_1=time | scriptable-transform --script="return ""#{payload.tr('^A-Za-z0-9', '')}""" --language=ruby | log
STREAM_2=:STREAM_1.time > scriptable-transform --script="function double(p) \n{\n     return p + '--' + p;\n}\ndouble(payload);" --language=javascript | log
STREAM_3=:STREAM_1.time > scriptable-transform --script="return payload + '::' + payload" --language=groovy | log
```

BoykoAlex:
Added Encoder and used it in stream service. Fixes #413

### DIFF
--- a/ui/src/app/shared/support/http.utils.spec.ts
+++ b/ui/src/app/shared/support/http.utils.spec.ts
@@ -1,0 +1,13 @@
+import { URL_QUERY_ENCODER } from './http.utils';
+
+describe('HTTP Utils', () => {
+
+  it('Encode + in key', () => {
+    expect(URL_QUERY_ENCODER.encodeKey('2 + 2 = 4')).toEqual('2%20%2B%202%20=%204');
+  });
+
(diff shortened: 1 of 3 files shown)


jvalkeal:
lgtm, tried it manually using pluses.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/d92742652a2e8c69aa9c78e0a93dba317704d4f1

PR #425, opened 2017-09-08
Changes Task Def Buttons to use simple icons.

### LINKED ISSUE #424: Switch task-definition action buttons with simple icons
As a developer, to be consistent, I'd like to switch buttons with simple `glyphicon` in task-definitions view. Currently, it is applied to Apps, Streams, Jobs, and Execution views.

cppwfs:
resolves #424


jvalkeal:
I'm going to polish that actions column width with merge.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/63a752a68abe1ea22964fc7a7eacb8524535e12f. I did some polish in a same way it was done for stream defs.

PR #442, opened 2017-09-11
Angular bootstrap tooltips for JointJS shapes

### LINKED ISSUE #438: Implement tool-tips

BoykoAlex:
Bootstrap tooltips for JointJS shapes via `ngx-bootstrap` lib.
Pivotal part of code substitutes regular Joint JS view markup rendering. Regular markup rendering is creation of a DOM from HTML markup string. It's substituted with DOM coming from an instantiated component that has the markup embedded in the Angular template. This allows the DOM to be controlled with angular components as well as allows angular directives to be applied to the DOM

Fixes #438


jvalkeal:
@BoykoAlex I think this PR looks ok. I didn't understand how this changes existing behaviour(is there any visible changes?) and I tried to compare UI to see if there's any changes to actual shown tool-tips.

BoykoAlex:
@jvalkeal while you hover over a shape in the palette are you not seeing a tooltip under the shape? Also once you drop the shape on a canvas and change it's properties the tooltip would show the set properties as well. When the shape in on the canvas ports would have tooltips, error markers would have have a red tooltips with a list of error messages. If a shape on the canvas is selected two handles show up at the bottom corners of a shape: remove element and edit properties. These two handles should have tooltips too. Are you seeing these tooltips? (I'll double-check my branch to see that this is the case)

jvalkeal:
Great, now I do see those. I wiped node_modules so maybe it fixed this. Merging.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/ad3c7c96c2abe60a68cfaf5c5674ddf3de317aa8.
I forget to amend author back to alex when I squashed those commits.

PR #449, opened 2017-09-12
Stream editor contents validation

### LINKED ISSUE #394: Compute DSL parse and graph validation errors together
Currently Flo validates the graph when graph changes and puts up markers on nodes/links right away. DSL is updated in async fashion but DSL parse errors are only computed when DSL text is modified by a user. Hence graph changes don't result in parse errors re-computation.

Need to compute parse errors and graph validation errors at the same time somehow. This will lead to cleaner `dsl-editor` component without previous hacks to update reconcile markers on DSL text. Also it'll allow to show up-to-date reconcile markers on DSL

### LINKED ISSUE #439: Add validation support for dsl-editor

BoykoAlex:
Adopt validation flow changes in Core Flo. Add `dsl-editor` validation rough-in. Parse errors should result as red squigglies in `dsl-editor`
Fixes #439 
Fixes #394

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -37,7 +37,7 @@
     "ngx-pagination": "^3.0.1",
     "rxjs": "^5.4.2",
     "sockjs-client": "^1.1.4",
-    "spring-flo": "git://github.com/spring-projects/spring-flo.git#b4e20eb9cd31764e2f96b178d03500c7d7d6628f",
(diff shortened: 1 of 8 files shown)


cppwfs:
LGTM.  The only thing I noticed (but not an issue with this code)
i could have a statement of foo=time and no error was displayed until I tried to create the definition.  I'll open that as a separate issue.

cppwfs:
Merged

PR #453, opened 2017-09-13
Validate against duplicate stream names on the form

### LINKED ISSUE #402: Validate duplicate streams
This feature was not brought over from old implementation because the implementation is hacky based on separate form control validation while it should have been done as a validator for the whole form group.

This validator should guard against duplicate stream names on the same form, i.e. stream on line 1 and stream on line 2 have same name situation.

BoykoAlex:
Validate against duplicate names on the form

Fixes #402 

If "Create stream dialog" has duplicate stream names on the from validation error should be displayed under the form controls having duplicate names

### DIFF
--- a/ui/src/app/streams/flo/support/utils.spec.ts
+++ b/ui/src/app/streams/flo/support/utils.spec.ts
@@ -139,4 +139,19 @@ describe('utils', () => {
     expect(Utils.generateStreamName(graph, http)).toEqual('unique-name');
   });
 
+  it('find duplicates: no elements', () => {
+    expect(Utils.findDuplicates([])).toEqual([]);
+  });
+
+  it('find duplicates: no duplicates', () => {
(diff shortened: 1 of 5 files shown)


jvalkeal:
@BoykoAlex Will this need a change to flo version
```
ERROR in /home/jvalkealahti/repos/spring/spring-cloud-dataflow-ui/ui/src/app/streams/flo/editor.service.ts (36,14): Class 'EditorService' incorrectly implements interface 'Editor'.
  Types of property 'validate' are incompatible.
    Type '(graph: Graph, dsl: string, flo: EditorContext) => Promise<Map<string, Marker[]>>' is not assignable to type '(graph: Graph) => Promise<Map<string, Marker[]>>'.
```

jvalkeal:
Never mind, looks like my npm modules were stale again. *sigh*

jvalkeal:
lgtm, merging.

Issue #455, opened 2017-09-13
Stream fails to deploy with security enabled


sabbyanandan:
As a user, I'm trying to deploy a stream (`time | log`), but I don't see it getting created/deployed when security is enabled. Everything works without security, though. This is the same behavior in local, cf, and k8s. 

To reproduce, 
> java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.3.0.BUILD-SNAPSHOT.jar --security.basic.enabled=true

Log in to the dashboard _(capture password from server logs)_ to create and deploy `time | log` stream. You will notice the stream validation messages in the logs, but the actual creation/deployment won't go through.

PR #466, opened 2017-09-14
gh-455 Fix: Stream fails to deploy with security enabled

ghillert:
- Ensure that we send the respective default headers:
  * `Content-Type`: `application/json`
  * `Accept`: `application/json`
- Fix tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/455


jvalkeal:
lgtm.

PR #467, opened 2017-09-14
Revisit validation

### LINKED ISSUE #395: Revisit validation routines in Flo
As a user, I'd like to see an incremental validation if there's more than 1 instance of the same named-channel in the canvas.

aclement:
EditorService
- Tests added (verifying validation of graph)
- Moved some messages to constants
- Altered validation to make full use of promises (rather than promises
  to a certain depth then async)
- More validation rules related to fan-in/out scenarios

graph-to-text
- Adjustments to cope better with 'bad input' and try to create some
  kind of graph (tests added for this case)

RenderService
- Prevents creation of tap links from destinations (channels/taps)
- Prevents link switching working for links from destinations

text-to-graph
- Fixed handling of map (use setter)

markup
- no bullet lists on error tooltips if only one entry
- tooltips working for destinations

Fixes #395

### DIFF
--- a/ui/src/app/streams/flo/decoration/decoration.component.html
+++ b/ui/src/app/streams/flo/decoration/decoration.component.html
@@ -6,7 +6,7 @@
 </svg:g>
 
 <template #markerTooltip>
-  <div>
+  <ul class="marker-tooltip">
     <li *ngFor="let msg of getMessages()">{{msg}}</li>
-  </div>
+  </ul>
 </template>
--- a/ui/src/app/streams/flo/decoration/decoration.component.scss
+++ b/ui/src/app/streams/flo/decoration/decoration.component.scss
@@ -1,7 +1,26 @@
 .error-marker-tooltip .tooltip-inner {
   background-color: red;
   max-width: 400px;
+  padding: 2px;
 }
+
 .tooltip.error-marker-tooltip .tooltip-arrow {
   border-right-color: red;
+}
+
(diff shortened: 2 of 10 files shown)


aclement:
For testing this you can fire up the Flo UI and try a few things:
- ensure you cannot create tap links from a destination
- Verify appropriate validation errors on bad graphs (build one with an unconnected source/processor/sink) - do the errors appear on the nodes correctly?

This does improve validation. However, it does not do what the description of 395 mentions - that is to put an error if you use a destination twice in a graph. This is not really an error, and will correct itself if you flip to working in the text view (the next graph generated won't include the dup). I'd like to decide how to properly treat this situation when working out what we do about loops.

The new fan-in/fan-out support gives a million new options for creating crazy graphs, I'm sure there are more issues to find as we do further testing.

jvalkeal:
@aclement we're getting merge conflict in src/app/streams/flo/render.service.ts prob with pr I just merged from alex

jvalkeal:
Ok, I'm fixing that conflict together with some lint errors. I probably can't review this as well as @sabbyanandan could but I poked around and think everything looks ok. So I prefer to merge this now and handle issues(if there's any) later.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/b4fd432a1342a132169195f2b74d1c6d6792ad5b

PR #468, opened 2017-09-14
Make rendering PhantomJS compatible to run complex tests

BoykoAlex:
Make Render service compatible with PhantomJS. Should allow for more complex tests. Made one test a bit more complex.


cppwfs:
LGTM

PR #470, opened 2017-09-14
Removes dashboard item poller when dashboard item is removed.

### LINKED ISSUE #460: Analytics: After clicking the reset button the timers are still firing for the displayItems that were removed.
This can be observed in the development console.

cppwfs:
resolves #460


ghillert:
LGTM merging.

Issue #459, opened 2017-09-13
Analytics: Adding a second counter resets name in first graph


cppwfs:
When adding a 2nd counter,  the first counter's `counter name` is reset.

PR #473, opened 2017-09-15
gh-459 Fix: Analytics - Adding a second counter resets name in first graph

ghillert:
* Make the counter collection part of each `DashboardItem`

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/459


cppwfs:
LGTM

PR #487, opened 2017-09-25
Initial infra for composed tasks

### LINKED ISSUE #484: Initial work for porting composed tasks
Work for this port is split into multiple smaller issues but until we get to those we need to add some base facilities and base code.

jvalkeal:
- Focus for this is to copy and implement tasks
  editors without touching any flo code in streams.
  Essentially this is to get dsl and flo editor to
  work on some level. Some error are still expected
  which will get fixed in other PR's.
- When tasks flo is closer to its full implementation,
  then it's easier to see what is the actual shared
  feature set in tasks and streams and extract that
  to shared classes. For this specific reason I'm
  referensing some stream flo classes from tasks.
- I'm starting to add tests in PR's after this.
- There has been some attempt to not use stuff under
  `joint.shapes.flo` directly as it will create situation
  where typescript compiles but may result runtime
  errors if some specific imports are missing.
- Fixes #484

### DIFF
--- a/ui/proxy.conf.json
+++ b/ui/proxy.conf.json
@@ -49,6 +49,11 @@
         "secure": false,
         "logLevel": "debug"
     },
+    "/tools" : {
+        "target": "http://localhost:9393/",
+        "secure": false,
+        "logLevel": "debug"
+    },
     "/runtime/apps" : {
         "target": "http://localhost:9393/",
         "secure": false,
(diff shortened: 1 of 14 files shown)


BoykoAlex:
LGTM as a base for further work. Let me know i should be merging this.

jvalkeal:
@BoykoAlex yes, it was just meant for having a base for further work.

BoykoAlex:
@jvalkeal Cool :-) @sabbyanandan i can merge it if you like. Feels like it might be blocking further work.

cppwfs:
LGTM as a base for CTR pages.

PR #489, opened 2017-09-27
Add tools service

### LINKED ISSUE #485: Add tools service
Flo is using server for some text/dsl conversions so i.e. `/tools/parseTaskTextToGraph` should be accessed via normal service class.

jvalkeal:
- Create new ToolsService class and domain classes
  accessing /tools to handle dsl-flo conversion logic.
  This is in favour of doing raw http calls in flo service
  classes.
- In flo metamodel/editor/render classes use this new
  service class.
- Refactor all the code around where data flows and we're
  now using those new domain classes instead of passing
  js objects around.
- Remove all dead and broken code around global properties
  in a flo. Essentially flo uses `timeout` property in start
  node together with rest of global properties. None of these
  exist in tasks nor in a service under `/tools`.
- Hard code group name matching as in old code it was done in
  matchGroup function which don't work anymore. Concept of that
  function made sense in streams space where you have more groups
  but here we have only one together with `control nodes`.
- Also all dead code which were brought over from streams to tasks
  (even in old code) are removed. Essentially this code we used in
  streams and had no meaning in tasks.
- Added broken popup for link properties. There is now a popup but
  its content view is broken as most of the features around
  node/cell properties are not yet implemented.
- Fixes #485

### DIFF
--- a/ui/src/app/tasks/flo/editor.service.ts
+++ b/ui/src/app/tasks/flo/editor.service.ts
@@ -8,6 +8,11 @@ import { PropertiesDialogComponent } from '../../streams/flo/properties/properti
 
 const joint: any = _joint;
 
+/**
+ * Flo service class for its Editor used for composed tasks.
+ *
+ * @author Janne Valkealahti
+ */
 @Injectable()
 export class EditorService implements Flo.Editor {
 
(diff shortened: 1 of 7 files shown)


sabbyanandan:
@BoykoAlex / @aclement: This PR removes _unused_ (based on Janne's interpretation) Flo code - please have a look and let us know what you think.

BoykoAlex:
I have added a tiny fix to check that only links with source and target are fed to graph->text conversion and then merged this with 2d359d627864d491d283604f111900951e429ff2

PR #493, opened 2017-10-02
Fix analytics polling stop

### LINKED ISSUE #418: Zero rate doesn't stop polling
In analytics counter if rate is set to 0 console logs
```
Changing refresh rate to ...0
```
but that's it. polling and new render still happens.

jvalkeal:
- In AnalyticsService check if given rate is undefined
  istead of simple `if(rate) as that will be false if
  zero is given.
- Add spec for AnalyticsService and test intervals.
- Imports polish for AnalyticsService
- Fixes #418


cppwfs:
LGTM.   Just threw me that we are not setting the counterInterval to zero.

PR #494, opened 2017-10-02
Support flo property validation against long or short names

### LINKED ISSUE #477: Incorrect validation on app properties
As a user, I'm trying to build a stream with an app and its associated properties, but the validation says it is bogus.

image

aclement:
Previously validation was only against long names (like
'trigger.time-unit' on the time module). With this change
validation also allows the short-form name 'time-unit' too.
Tests added.

Fixes #477


BoykoAlex:
LGTM. Merged with minor fix-up with 168d810df84c03d4d9f93893070d1122956e13c9

Issue #325, opened 2017-08-03
Validation not working for bulk task definitions


sabbyanandan:
As a user, I'm trying to define tasks in bulk, but the duplicate definitions are not validated. They are silently ignored. The same is also true when an invalid task is used in the definition - no error - silently ignored again. 

![image](https://user-images.githubusercontent.com/1562654/28943153-8ed60e4c-7852-11e7-81c4-533e776a7e03.png)

sabbyanandan:
I see the following in the console. The error message doesn't show up in the UI, though.

![image](https://user-images.githubusercontent.com/1562654/28944408-3cd429c6-7857-11e7-93ba-dfe61f6e692b.png)

sabbyanandan:
This is addressed via: https://github.com/spring-cloud/spring-cloud-dataflow-ui/commit/8edffdb32d8841e0cfbe485181d961c7905e3b94.

PR #498, opened 2017-10-03
gh-325 Fix: Validation not working for bulk task definitions

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/325


BoykoAlex:
LGTM. I can merge.
I have one remark. If I type task def: `mytask1=timestamp --p1=value`. It's hsowing up as invalid since `Create` button is disabled - think it should be a valid definition in terms of parser. Yet when i type `mytask1=timestamp --p1` it's shown as valid, `Create` button enabled... @ghillert Another bug?

BoykoAlex:
Merged with 8edffdb32d8841e0cfbe485181d961c7905e3b94

PR #501, opened 2017-10-03
Create Stream button enablement based on validation

### LINKED ISSUE #450: Disable "create stream" button if there are validation errors in the DSL/graph
`foo=time` will not show an error with syntax highlighting.  It is caught however, when attempting to creae stream.

BoykoAlex:
* `Create Stream` button enablement based on presence validation error markers and parse errors
* Fixed an issue with property stream having a long name property but properties dialog setting the short name property in this case. Should just update the long name property if it's being changed

Fixes #450

### DIFF
--- a/ui/src/app/streams/flo/properties/properties-dialog.component.ts
+++ b/ui/src/app/streams/flo/properties/properties-dialog.component.ts
@@ -100,14 +100,29 @@ class PropertiesGroupModel extends Properties.PropertiesGroupModel {
     });
   }
 
+  private determineAttributeName(metadata: Flo.PropertyMetadata): string {
+    const nameAttr = `props/${metadata.name}`;
+    const idAttr = `props/${metadata.id}`;
+    if (this.cell.attr('metadata/group') === 'other') {
+      // For something in the other group (like tap) use the id not the name of the property
+      return idAttr;
+    }
+    const valueFromName = this.cell.attr(nameAttr);
+    const valueFromId = this.cell.attr(idAttr);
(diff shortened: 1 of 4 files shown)


ghillert:
@BoykoAlex looks mostly good - There is 1 little UX issue. When I have an empty text-area and I start to type the first letter, then the `Create Stream` button is enabled for a split-second.

image

I was wondering, if you could address that before merging.

BoykoAlex:
@ghillert can we raise that a separate issue? I won't be able to address this right away unfortunately (tied up on other project), but am interested to look at a few options to address this perhaps.
The issue is validation kicks in after a debounce timeout time interval after user stops typing. There is no ideal fix for this. I see 2 options:
1. Make that timeout shorter (if CodeMirror leaves anything for the value to be customized)
2. Disable the button until validation is finished (for this I'd need to know when contents became invalid - an observer from Flo). This will make validation button ti flicker while graph/text being edited.

BoykoAlex:
@ghillert I'm in favour of pushing this and then i get back to what's left whenever the next time I get a chance to look at SCDF UI/Flo stuff

ghillert:
@BoykoAlex @sabbyanandan Sounds good. I have created https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/512 as a follow-up.

ghillert:
LGTM - merging.

ghillert:
merged with commit deae905f3535fa297179a9e28d9e4574c2ff5dd2

Issue #491, opened 2017-10-02
Analytics dashboard selectors don't restore


jvalkeal:
When going between dashboard and counters, counter name and visualization drop-downs doesn't restory while metric type and graph itself are restored.

jvalkeal:
Also only once I saw below errors in a console when I was poking this issues. I couldn't re-produce it but may be something to do with how chart and its data calls are restored. As switchmap is used in a chain to replace old subscription then something probably went wrong as there was this `undefined`.  

```
error TypeError: You provided 'undefined' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
    at Object.subscribeToResult (subscribeToResult.js:73)
    at SwitchMapSubscriber.webpackJsonp.../../../../rxjs/operator/switchMap.js.SwitchMapSubscriber._innerSub (switchMap.js:101)
    at SwitchMapSubscriber.webpackJsonp.../../../../rxjs/operator/switchMap.js.SwitchMapSubscriber._next (switchMap.js:94)
    at SwitchMapSubscriber.webpackJsonp.../../../../rxjs/Subscriber.js.Subscriber.next (Subscriber.js:89)
    at AsyncAction.webpackJsonp.../../../../rxjs/observable/IntervalObservable.js.IntervalObservable.dispatch [as work] (IntervalObservable.js:70)
    at AsyncAction.webpackJsonp.../../../../rxjs/scheduler/AsyncAction.js.AsyncAction._execute (AsyncAction.js:111)
    at AsyncAction.webpackJsonp.../../../../rxjs/scheduler/AsyncAction.js.AsyncAction.execute (AsyncAction.js:86)
    at AsyncScheduler.webpackJsonp.../../../../rxjs/scheduler/AsyncScheduler.js.AsyncScheduler.flush (AsyncScheduler.js:36)
    at ZoneDelegate.webpackJsonp.../../../../zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:425)
    at Object.onInvokeTask (core.es5.js:3881)
```

PR #504, opened 2017-10-04
gh-491 Analytics dashboard selectors don't restore

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/491

### DIFF
--- a/ui/src/app/analytics/analytics.service.ts
+++ b/ui/src/app/analytics/analytics.service.ts
@@ -359,7 +359,7 @@ export class AnalyticsService {
    */
   public restartPollingOfSingleDashboardItem(dashboardItem: DashboardItem) {
     this.stopPollingOfSingleDashboardItem(dashboardItem);
-    if (dashboardItem.refreshRate > 0 ) {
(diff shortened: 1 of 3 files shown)


cppwfs:
Found one issue.  If a user selects just a metric type(nothing else) and then clicks the job tab and then returns back to the analytics page, the page is `busy` and both the analytics and job components are showing.   The error displayed is as follows: 
```
ERROR TypeError: Cannot read property 'name' of undefined
    at AnalyticsService.webpackJsonp.../../../../../src/app/analytics/analytics.service.ts.AnalyticsService.startPollingForSingleDashboardItem (analytics.service.ts:305)
```

ghillert:
@cppwfs 

- Rebased
- Improved error handling (generate toasty message in case Redis is not running for counters)
- Made additional check when starting the polling for a dashboard item (Ensure Counter is defined)

cppwfs:
LGTM WOOT! WOOT!

cppwfs:
Squashed Merged

Issue #499, opened 2017-10-03
Provide better messaging if no Apps are registered (UI regression 1.2.3 -> 1.3.0.M2)


sabbyanandan:
_From @pawelantczak on September 28, 2017 10:46_

Hello.

While on ```1.2.3``` UI looks and works like it should,
on ```1.3.0.M1``` some resources (styles?) are missing causing UI rendering issues.
![screenshot 2017-09-28 12 40 09](https://user-images.githubusercontent.com/171534/30962641-d7c5bf3e-a44a-11e7-9cd6-a166a35f794f.png)
![screenshot 2017-09-28 12 39 04](https://user-images.githubusercontent.com/171534/30962645-dca50ef6-a44a-11e7-9a96-c62076c77923.png)
 

_Copied from original issue: spring-cloud/spring-cloud-dataflow#1666_

ghillert:
The solution will render a message similar to the old UI. In addition to the old UI, the new approach will also be `role` aware. So if you only have a `VIEW` role, the following screen will be shown:

![image](https://user-images.githubusercontent.com/685363/31197802-f135ce9e-a8ed-11e7-84de-2ed6b4d40962.png)

However, if you have the `CREATE` role, the complete message is shown:

![image](https://user-images.githubusercontent.com/685363/31197880-34d06100-a8ee-11e7-8d4d-bad5c3ad69d7.png)

pawelantczak:
Hello.
Please take a look at styling, top menu for example.
It was not about message, the whole UI if affected by missing js/css a believe.
Regards.

sabbyanandan:
Hi, @pawelantczak: we are rewriting the Dashboard entirely in the 1.3 line. I believe you're referring to the newer look & feel - please refer to the [1.3 M1 release blog](https://spring.io/blog/2017/08/07/spring-cloud-data-flow-1-3-0-m1-released#angular-4) for more details.

PR #505, opened 2017-10-04
gh-499 Provide better messaging if no Apps are registered

ghillert:
* Render a message similar to the old UI.
* Make the message role aware
  - `Register Application(s)` button is only shown if user has the `CREATE` role

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/499

See also screenshots added under https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/499


cppwfs:
LGTM

PR #510, opened 2017-10-11
Add composed task create dialog

### LINKED ISSUE #445: Port composed-task to Angular4 infrastructure
As a developer, I'd like to refactor and port composed-task canvas with Angular4 stack.

jvalkeal:
- Fixed bug which did add link in flo to a structure
  thus graph to text was broken.
- Add similar type of dialog used in streams to
  submit task dsl with its name.
- Fixes #445

### DIFF
--- a/ui/src/app/tasks/flo/metamodel.service.ts
+++ b/ui/src/app/tasks/flo/metamodel.service.ts
@@ -206,7 +206,7 @@ export class MetamodelService implements Flo.Metamodel {
           metadata.set(this.COMPOSED_TASK_LABEL, element.attr('node-label'));
         }
         nodes.push(new Node(element.attributes.id, attrs.metadata.name, element.attributes.attrs.props, metadata));
-      } else if (element.get('type') === joint.shapes.flo.LINK_TYPE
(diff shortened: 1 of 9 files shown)


ui/src/app/tasks/flo/metamodel.service.ts
```diff
@@ -206,7 +206,7 @@ export class MetamodelService implements Flo.Metamodel {
           metadata.set(this.COMPOSED_TASK_LABEL, element.attr('node-label'));
         }
         nodes.push(new Node(element.attributes.id, attrs.metadata.name, element.attributes.attrs.props, metadata));
-      } else if (element.get('type') === joint.shapes.flo.LINK_TYPE
+      } else if ((element.get('type') === joint.shapes.flo.LINK_TYPE || element.attributes.type === 'link')
```
BoykoAlex: Hmm... was that the fix for graph - to - text? Don't think the type of the link should be anything else but ` joint.shapes.flo.LINK_TYPE`. Don't think that `'link'` should be there... wonder where it comes from. @jvalkeal please confirm if this is intentional change. If yes, I'll merge and check what lead to type being `'link'`

BoykoAlex:
Merged with 1d35c9b63a3d36e3da0eb003326f286f9b310351

PR #530, opened 2017-11-03
Task defs list up to date after task create

### LINKED ISSUE #476: Reload task list on redirect
As a user, I create a task from the dashboard and that results in a redirect to the list page. However, the results are stale and it'd be nice to fetch the latest on redirect.

It appears to be a problem when there's only one task - worth double checking.

cppwfs:
resolves #476

This was a sporadic issue that only occured when using the H2 DB.
And could not be reproduced using the server provided by angular.
The source problem was that the redirect only navigate was outside
the Observable and it would fire before the observable was complete.

To reproduce the problem.
. Build UI using `mvn clean install`
. Update SCDF pom to use BUILD-SNAPSHOT of UI
. Build SCDF
. Start SCDF using the default H2 database
. Add task definitions.


ghillert:
Merged with commit 582a9e365a3032dba3ad7ca7f1db002f3748ec8d

PR #532, opened 2017-11-03
refreshes task definitions on delete

### LINKED ISSUE #531: CTR child task defs remain on def list page after delete
When deleting a CTR task definition using the UI the system goes back to the def list page.  The CTR child task definitions are still present.  Upon refresh of the page they no longer show up.  
The root issue is how the page handles the deletion of the task definition.   As shown here:
https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/master/ui/src/app/tasks/tasks.service.ts#L201.   Only the CTR task Definition is removed from the cached response instead of the refreshing the page or removing the child tasks from the cache.

cppwfs:
resolves #531


ghillert:
Merged with commit 7b6e293f31ea7bd3284e2bf61df079c81e81d6c5

PR #537, opened 2017-11-07
Allow app and deployment properties to have more than 1 = char

### LINKED ISSUE #529: Validation is not allowing environmentVariables
Hey there. It appears that I cannot add a property like the one below when deploying a stream:

deployer.*.kubernetes.environmentVariables="JAVA_TOOL_OPTIONS=-Xmx384m"

The problem appears to be [here](https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/b41af247de589df1273c101a0f4f3bacd796e71f/ui/src/app/streams/stream-deploy/stream-deploy-validators.ts#L17)

cppwfs:
resolves #529


ghillert:
LGTM merging

ghillert:
Merged with commit 95e9a0ee7d261e8a480eaeae2111a68758f1dc45

PR #553, opened 2017-11-16
Fix default boolean properties for the dialog

### LINKED ISSUE #552: App properties dialog shows wrong defaults for `true` boolean properties
1. Drop `file` source on the Flo editor (create stream page)
2. Select it, click on the "cog" icon to open properties dialog
3. Note `prevent-duplicates` and `markers-josn` properties. they are unchecked, but should be checked because ther default values are `true`
Now if you try to set/unset the boolean value for these properties you'll realize that it's just broken as a consequence of this issue.

BoykoAlex:
Fixes #552

### DIFF
--- a/ui/src/app/shared/flo/support/properties-group-model.ts
+++ b/ui/src/app/shared/flo/support/properties-group-model.ts
@@ -29,8 +29,7 @@ export class PropertiesGroupModel extends Properties.PropertiesGroupModel {
           inputType = Properties.InputType.URL;
           break;
         case 'java.lang.Boolean':
-          inputType = Properties.InputType.CHECKBOX;
-          break;
+          return new Properties.CheckBoxControlModel(property);
         default:
           if (property.metadata.code) {
             if (property.metadata.code.langPropertyName) {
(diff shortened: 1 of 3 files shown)


BoykoAlex:
@ghillert @sabbyanandan it'd be nice to get this PR in. It fixes an small but rather important issue as well as bumps up Flo to `0.7.0-beta` avoiding completely the `postinstall` stuff as Flo only published the built artifacts. The VersionEye still shows no license, but perhaps it needs time to refresh caches... not sure... Anyway if we get it in we have a great opportunity to find out whether the VersionEye complaint is fixed.

ui/src/app/shared/flo/support/properties-group-model.ts
```diff
@@ -42,9 +41,9 @@ export class PropertiesGroupModel extends Properties.PropertiesGroupModel {
             }
           } else if (Array.isArray(property.metadata.options)) {
             return new Properties.SelectControlModel(property,
-              Properties.InputType.SELECT, (<Array<string>> property.metadata.options).map(o => {
+              Properties.InputType.SELECT, (<Array<string>> property.metadata.options).filter(o => o.length > 0).map(o => {
```
ghillert: I will merge BUT this feels like code-smell. We really need to have proper Java-style enum-support across the app. I wish Typescript was a bit better in that regard. Maybe we should consider using `ts-enums` --> https://github.com/LMFinney/ts-enums

PR #558, opened 2017-11-28
UI now supports checksum for data flow components

### LINKED ISSUE #556: About tab should display checksum information returned from dataflow
Tab should also use the shell information returned as a part of the result.
Tab also displays the following error message `Error occurred when attempting to obtain about info from server` while waiting for the result.  This should be removed..

cppwfs:
resolves #556


ghillert:
LGTM - merging

Issue #559, opened 2017-11-28
Tabs don't show up after authentication


sabbyanandan:
As a user, I gain access to the dashboard after authentication; however, the tabs don't show up. It does when I manually reload the page, though.

before:

![image](https://user-images.githubusercontent.com/1562654/33324849-9f379396-d405-11e7-8084-7a2131be270e.png)

after refresh:

![image](https://user-images.githubusercontent.com/1562654/33324825-8d1bb25a-d405-11e7-8732-8e27706a08c2.png)

PR #561, opened 2017-11-29
gh-559 Fix: Tabs don't show up after authentication

ghillert:
- Ensure that roles are checked after login
- Capture pre-existing `display` css property (used for reset)
- Use a `Subject` instead of relying on `ngDoCheck` to react to role-changes

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/559


cppwfs:
LGTM

cppwfs:
Squashed Merged.

PR #567, opened 2017-12-11
Off page rendering for fitLabel(...) function implementation

BoykoAlex:
A lot of white space changes due to IntelliJ `Code Format` - think it's good actually.
`fitLabel(...)` function is reworked to measure `tspan` size created off page with the same styling that is present on the actual page. Speeds up palette rendering by 20-25% (instead of updating and measuring text continuously on the page)

### DIFF
--- a/ui/src/app/streams/flo/render.service.ts
+++ b/ui/src/app/streams/flo/render.service.ts
@@ -15,60 +15,19 @@
  */
 
 import './support/shapes';
-import { IMAGE_W, HORIZONTAL_PADDING } from './support/shapes';
 import {
-  Injectable, ComponentFactoryResolver, ComponentFactory, Injector, ApplicationRef, Type,
-  ComponentRef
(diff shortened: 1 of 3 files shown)


sabbyanandan:
@ghillert: Please have a look when you get a chance.

ghillert:
@BoykoAlex There is a lint error:

`render.service.ts[133, 15]: Shadowed variable: 'textView'`

ghillert:
@BoykoAlex Not a deal maker for now - But do you think we could maybe chop this class into smaller pieces? Almost a 1000 lines of code is pushing it a bit IMO. Could you maybe create a follow-up issue?

BoykoAlex:
@ghillert Fixed lint erro and factored out two helper classes from `Renderer` class.

ghillert:
Merged with commit d1f6c620c83f03ec6fcdc74bc85c1afe09ab8621

PR #583, opened 2018-01-02
Change the no task definition message on jobs

### LINKED ISSUE #579: Tasks run externally from dataflow show `DELETED` in the UI.  Should read as `No Task Definition` instead.

cppwfs:
Change class from Danger to Info
Updated message from `Deleted` to `No Task Definition`.

resolves #579


ghillert:
LGTM

PR #586, opened 2018-01-09
Override link options to prevent link tools resizing

### LINKED ISSUE #393: Primary/secondary cog position isn't too friendly
As a user, I'm creating a complex graph and for topologies involving fan-out, I cann't easily select primary/secondary cog. It seems to be fixed to a position where it is already too crowded.

aclement:
All options are the same as default except shortLinkLength which has been reduced to 0. Normally if link length is < shortLinkLength then the size of the tools is halved.

resolves #393

### DIFF
--- a/ui/src/app/streams/flo/support/view-helper.ts
+++ b/ui/src/app/streams/flo/support/view-helper.ts
@@ -61,7 +61,14 @@ export class ViewHelper {
     const V = joint.V;
 
     return joint.shapes.flo.LinkView.extend({
-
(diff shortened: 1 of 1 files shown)


sabbyanandan:
The buttons look so much easier to operate on. Thanks for fixing it! 

image

@BoykoAlex: Could you review and merge it, please?

BoykoAlex:
LGTM. Merged d442f0d64475994cf4105a00e20b60f150f8226b

PR #589, opened 2018-01-09
Hide task-launch button for VIEW-only users

sabbyanandan:
See the [comment](https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/588#issuecomment-356436316) for background and the proposed solution. 

This was verified with a user having a VIEW-only role. Likewise, when the role is switched to CREATE, all the buttons will show.

Resolves spring-cloud/spring-cloud-dataflow-ui#588.


cppwfs:
Rebased, Merged

PR #604, opened 2018-01-24
Ensure recognize tap streams when computing text from graph

### LINKED ISSUE #603: Switching links between 'regular' and 'tap' in create streams page can misbehave
Create the stream `http | cassandra`, now drag out a new `file` sync and join it to http, the streams will be correct (`STREAM_1 = http | cassandra` and `:STREAM_1.http > file`).  Now if you click the 'flip link' on each of the links to make `http` to `file` the primary and `cassandra` the tap, the DSL text will not be correct.

aclement:
Without this change if the first stream in a graph is a tap stream
the graph to text algorithm will make a mistake and not treat it
as a tap stream because Array.find returns 0 in this case (because
it is stream 0 that is a tap stream). 0 is treated as false. It
is better to check the Array.find call returned a result (i.e.
result is not undefined)

Fixes #603

### DIFF
--- a/ui/src/app/streams/flo/graph-to-text.spec.ts
+++ b/ui/src/app/streams/flo/graph-to-text.spec.ts
@@ -398,6 +398,16 @@ describe('graph-to-text', () => {
         expect(dsl).toEqual('aaa=time | logA\n:aaa.time > logB');
     });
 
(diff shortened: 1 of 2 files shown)


cppwfs:
LGTM

Issue #574, opened 2017-12-20
Attempting to destroy a task in Safari browser shows a black-screen overlay


sabbyanandan:
notes from @trisberg:

```
if I hit `Esc` it briefly shows the dialog and then it goes away

looks like there is some kind of overlay that gets in the way
``` 
example:

![image](https://user-images.githubusercontent.com/1562654/34224251-210a7672-e577-11e7-9ad1-c087e03e65d1.png)

PR #609, opened 2018-01-29
gh-574 Fix Safari/FF browser overlay issue

ghillert:
* Remove .fade css animation from styles (main cause of issue)
* Add web-animations-js polyfill so that web animations work in Safari
  - needed by the busy module so that fade-ins/outs still work
* Fixes the issues in Safari and Firefox

fixes https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/574


cppwfs:
LGTM

PR #619, opened 2018-02-08
Runtime Module: coverage tests, documentation, optimizations

oodamien:
Resolves spring-cloud/spring-cloud-dataflow-ui#543


ui/src/app/tests/mocks/modal.ts
```diff
@@ -15,5 +15,16 @@ import {TemplateRef} from '@angular/core';
  * @author Damien Vitrac
  */
 export class MockBsModalRef extends BsModalRef {
-  hide = Function
+
+  hide() {};
```
ghillert: Compilation error - this works though `hide = Function;`


```
ERROR in src/app/tests/mocks/modal.ts(19,3): error TS2425: Class 'BsModalRef' defines instance member property 'hide', but extended class 'MockBsModalRef' defines it as instance member function.
```

ui/src/app/runtime/runtime-apps.service.ts
```diff
@@ -5,55 +5,63 @@ import 'rxjs/add/operator/catch';
...
-            .map(this.extractData.bind(this))
-            .catch(this.errorHandler.handleError);
-    }
+  /**
+   * Get the runtime applications based on the paged parameters.
+   *
+   * @returns {Observable<R|T>} the Promise that returns the paged result of Runtime applications.
+   * @param pagination
+   */
+  public getRuntimeApps(pagination: { page: number, size: number }): Observable<Page<RuntimeApp>> {
```
ghillert: I am a little hesitant - Would maybe a `PageInfo` defined as interface work? I think that then you can just pass in a JSON object as parameter as long as it adheres to the *Interface*.

I kind of like having a well-defined interface/class here as I can then introspect the app and ask question like - show me all the methods in the app that perform pagination and pagination itself follows a certain contract.

ui/src/app/runtime/runtime-apps/runtime-apps.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Subscription of Runtime Apps
+   */
+  subscriptionRuntimeApps: Subscription;
+
+  /**
+   * Pagination state
+   * @type {{page: number; size: number}}
+   */
+  pagination = {page: 0, size: 10};
```
ghillert: See my other comment - I think this is too generic and would like something more typed - I like concrete classes but in this case an Interface should be fine, e.g:

```
export interface Pagination {
  page: number;
  size: number;
} 
const pagination: Pagination = { page:1, size:10 }
```

Then you can at least say: `pagination: Pagination = {page: 0, size: 10}; `

(And if somebody fatfingers something else in between the curly braces, the compiler will at least complain.)

oodamien: Yes. The interface is in the PR 612 (Pagination, Filter, Sort Interface).

ui/src/app/runtime/runtime-apps.service.ts
oodamien: Yes. The interface is in the PR 613 (Pagination, Filter, Sort Interface).

ghillert:
Merged with commit af548159398e384ab208590bb8004d4cf422b310

PR #627, opened 2018-02-12
[DO NOT MERGE] Stream definitions: add constraints on deployment properties

### LINKED ISSUE #515: Automate stream deployment properties based on platform selection
When deploying a single stream or if bulk-deploying a stream, you can currently specify ANY property key, when specifying deployment properties. This will error out on the server-side with:

```
Only deployment property keys starting with 'app.' or 'deployer.' allowed, got 'foo.bar'
```

We should improve the validator to pro-actively only allow for those 2 prefixes.

oodamien:
Resolve #515
The key must start with `app.` or `deployer.`.
The key must contain only alphanumeric characters.


sabbyanandan:
This PR will be closed in favor of stream-deployment page redesign.

Issue #615, opened 2018-02-06
Bogus SVG elements are appended to the body


oodamien:
When I navigate on the stream creation page, SVG elements are append to the `body`.

![image](https://user-images.githubusercontent.com/487067/35867850-21011b0e-0b5b-11e8-845b-b30bcf33e74f.png)

This elements persist even after navigate to an another page.

Following, the SVG elements display ('opacity' set to 1)

![image](https://user-images.githubusercontent.com/487067/35867975-8c69d4ee-0b5b-11e8-80d6-d1c74d28c3e5.png)

BoykoAlex:
Looks like `spring-flo` bug. Most likely DnD from palette adds shapes to body as "feedback" shapes during DnD from the palette to the canvas. Good catch - needs confirmation and then fixing.

PR #639, opened 2018-02-16
Use ApplicationType in SharedAppService#getApps() rather than `any` type

### LINKED ISSUE #638: Fix Composed Task Creation page throws error
Looks like the application type is not resolved correctly:

image

BoykoAlex:
Fixes #638 
Adjust API to use `ApplicationType` enum type rather than `any` type to match signature of `#getAppInfo(...)`
Add fix for #615 to Tasks rendering


ghillert:
LGTM - merging

PR #674, opened 2018-03-09
Stream Deployment: redirect if stream not exist

### LINKED ISSUE #662: Improve UX when deploying non-existing stream
**Depends on** https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/755

When deploying a non-existing stream, e.g. `http://localhost:4200/#/streams/definitions/test/deploy` the UI looks like:

image

We should instead redirect to the streams page with a notification message stating that "the stream does not exist".

oodamien:
Resolves #662

### DIFF
--- a/ui/src/app/streams/stream-deploy/stream-deploy.component.ts
+++ b/ui/src/app/streams/stream-deploy/stream-deploy.component.ts
@@ -5,7 +5,7 @@ import {Subscription} from 'rxjs/Subscription';
 import {Observable} from 'rxjs/Observable';
 import {StreamDeployService} from './stream-deploy.service';
 import {StreamDefinition} from '../model/stream-definition';
-import {map, takeUntil} from 'rxjs/operators';
+import {catchError, map, takeUntil} from 'rxjs/operators';
 import {StreamDeployValidator} from './stream-deploy.validator';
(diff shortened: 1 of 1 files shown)


ui/src/app/streams/stream-deploy/stream-deploy.component.ts
```diff
@@ -88,6 +88,12 @@ export class StreamDeployComponent implements OnInit, OnDestroy {
     this.route.params.subscribe((params: Params) => {
       this.config$ = this.streamDeployService
         .config(params.id)
+        .pipe(catchError(val => {
+          if (val === `Could not find stream definition named ${params.id}`) {
```
ghillert: Not sure - I like this.

Maybe we should defer to `1.5` and handle/solve this more appropriately. Our current error handler is somewhat limited - returning a simple string has not been the best thing _since sliced bread_. We should either add a generic `NotFoundException` that can be caught by component code or we treat a `404` status specifically, causing business methods to return `undefined` in that case.

ghillert:
@sabbyanandan Closing for now. The main issue can be left open but would depend on https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/755

PR #707, opened 2018-03-30
Stream list: wording

### LINKED ISSUE #706: No registered apps - Don't show Create Stream button
When going to `http://localhost:4200/#/streams` and no Apps are registered the current messaging makes no sense. There should not be a `Create a Stream` button, but instead a way to redirect the user to the Apps screen to register new Apps.

image

oodamien:
Resolves #706


ghillert:
Merged with commit 70e7da23d8e6a9ad478b76e734a944dc0ba8c74a

PR #744, opened 2018-04-24
Task execution: fix URL

### LINKED ISSUE #741: Broken link from task execution to batch execution details
Hi
The link from task execution to batch execution is broken, it  uses wrong context,
should be `dashboard/#/jobs/executions/1009`
image

oodamien:
Resolves #741

### DIFF
--- a/ui/src/app/tasks/task-execution/task-execution.component.html
+++ b/ui/src/app/tasks/task-execution/task-execution.component.html
@@ -50,7 +50,7 @@
     <tr>
       <td>Job Execution Ids</td>
       <td *ngIf="taskExecution.jobExecutionIds.length > 0" id="jobExecutionIds">
-        <a class="btn btn-default" type="button" [routerLink]="'jobs/executions/' + jobExecutionId"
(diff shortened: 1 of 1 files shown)


ui/src/app/tasks/task-execution/task-execution.component.html
```diff
@@ -50,7 +50,7 @@ <h1>Task Execution Details - Execution ID: {{ taskExecution.executionId }}</h1>
     <tr>
       <td>Job Execution Ids</td>
       <td *ngIf="taskExecution.jobExecutionIds.length > 0" id="jobExecutionIds">
-        <a class="btn btn-default" type="button" [routerLink]="'jobs/executions/' + jobExecutionId"
```
ilayaperumalg: While the routerLink is fixed now, don't we need the style for the link(s) still?

oodamien: The button render is not really nice in a table, we have to use classic link in that case.

PR #754, opened 2018-04-26
Error for multiline CTR definitions. Workarounds for parsing empty DSL

### LINKED ISSUE #723: Disallow multiple task definitions in text-area
Currently we allow multiple lines in the textarea for the Task Definition editor. This can cause the Flo editor to get out-of-sync:

image

image

- We should disallow multiline definitions. 
- If the user adds multiple lines we should show an error AND disable the `CREATE TASK` button

Currently users can click the `Create Task` button:

image

And then clicking the `Create` button gives you a vaguely cryptic error message.

BoykoAlex:
Fixes #723

### DIFF
--- a/ui/src/app/tasks/task-definition-create/task-definition-create.component.ts
+++ b/ui/src/app/tasks/task-definition-create/task-definition-create.component.ts
@@ -28,8 +28,10 @@ export class TaskDefinitionCreateComponent implements OnInit, OnDestroy {
   paletteSize = 170;
   editorContext: Flo.EditorContext;
 
+  contentValidated = false;
   lintOptions: CodeMirror.LintOptions;
   validationMarkers: Map<string, Flo.Marker[]>;
+  parseErrors: any[];
 
   initSubject: Subject<void>;
   busy: Subscription;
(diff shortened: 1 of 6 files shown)


ghillert:
@BoykoAlex  If you can, please disable the `Create Task` button in case of a multi-line Task definition:

image

ui/src/app/tasks/task-definition-create/task-definition-create.component.ts
```diff
@@ -107,34 +107,36 @@ export class TaskDefinitionCreateComponent implements OnInit, OnDestroy {
...
-            severity: 'error'
-          }));
-        }
+      const doc = editor.getDoc();
+      if (doc.lineCount() > 1) {
+        // Multiline definitions are not supported. Parser throws errors
+        annotations.push({
+          from: doc.posFromIndex(0),
+          to: doc.posFromIndex(doc.getValue().length - 1),
+          message: 'Multi-line composed task definitions are not supported',
```
ghillert: Change to `Multi-line task definitions are not supported`. Since the former `Composed Task` editor is now **THE** editor for ALL tasks, we probably should not use the term `composed` in too many contexts.

BoykoAlex:
@ghillert corrected the string, added proper disabling of the `Create` button similar to Streams... Reworked the fix for multi-line a bit...

ghillert:
LGTM. Polished some minor lint errors (missing semicolons). Merged with commit 6ccea4e0ce300996002b5521021424a56e14b35f

PR #764, opened 2018-05-03
Back action in child pages

### LINKED ISSUE #699: What does `cancel` mean?
We start having more and more cases where we can cross-reference component state. E.g. a task execution may have a link to job executions but we also have a job execution list page links to a job execution details page. Therefore, a page may have been entered from multiple source locations. Therefore, the user might have different expectation of where (s)he is expected to be routed when pressing a `Cancel` button. I wonder if we may have to have some form of `state management service` that keeps track of where the user might have been. So when somebody presses `Cancel`, then we may have a better idea of where to route the user to.

oodamien:
Resolves #699


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged into master as 0779664 and cherry-picked, pushed into 1.5.x as c4ed954

PR #765, opened 2018-05-03
HTTP error / ErrorHandler improvements

### LINKED ISSUE #662: Improve UX when deploying non-existing stream

### LINKED ISSUE #755: Make `ErrorHandler` or `general error handling` smarter
The `ErrorHandler` currently only returns a `string`. We need to make the `ErrorHandler` (
https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/master/ui/src/app/shared/model/error-handler.ts) smarter in a sense that we can differentiate:

* Errors that can be handled by the application (Expected errors)
* And fatal/errors that cannot be handled (Unexpected errors)

Our current error handler is somewhat limited - returning a simple string has not been the best thing. The main use-case is for REST calls that return a `404`. 

Alternatively, we should either add a generic `NotFoundException` that can be caught by component code or we treat a `404` status specifically, causing business methods to return undefined in that case.

oodamien:
Resolves #755
Resolves #662

### DIFF
--- a/ui/src/app/jobs/job-execution-details/job-execution-details.component.ts
+++ b/ui/src/app/jobs/job-execution-details/job-execution-details.component.ts
@@ -6,6 +6,9 @@ import { StepExecution } from '../model/step-execution.model';
 import { mergeMap } from 'rxjs/operators';
 import { Observable } from 'rxjs/Observable';
 import { NotificationService } from '../../shared/services/notification.service';
+import { AppError, HttpAppError } from '../../shared/model/error.model';
+import { RoutingStateService } from '../../shared/services/routing-state.service';
+import { LoggerService } from '../../shared/services/logger.service';
 
 /**
  * Displays a job's execution detail information based on the job execution id that is passed in via params on the URI.
@@ -29,13 +32,19 @@ export class JobExecutionDetailsComponent implements OnInit {
    * Constructor
    *
    * @param {JobsService} jobsService
+   * @param {NotificationService} notificationService
    * @param {ActivatedRoute} route
+   * @param {LoggerService} loggerService
+   * @param {RoutingStateService} routingStateService
    * @param {Router} router
    */
   constructor(private jobsService: JobsService,
               private notificationService: NotificationService,
               private route: ActivatedRoute,
+              private loggerService: LoggerService,
+              private routingStateService: RoutingStateService,
               private router: Router) {
+
   }
 
   /**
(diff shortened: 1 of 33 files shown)


ui/src/app/shared/model/error.model.ts
```diff
@@ -0,0 +1,91 @@
...
+  constructor(message: string) {
+    this.message = message;
+  }
+
+  /**
+   * ToString
+   *
+   * @returns {string}
+   */
+  toString() {
```
ghillert: I would like to rename that to `getMessage()` - that way it feels a bit more similar to Java :-) Since coming from that world `toString()` may not imply a end-user-destined message but may contain other debug-info.

ui/src/app/shared/model/error.model.ts
```diff
@@ -0,0 +1,91 @@
...
+   * Status code
+   */
+  status: number;
+
+  /**
+   * Constructor
+   * @param {string} message
+   * @param {number} status
+   */
+  constructor(message: string, status: number) {
```
ghillert: rename `status` to `httpStatusCode`

ui/src/app/shared/model/error.model.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * API Error
+ */
+export class AppErrorApi extends AppError {
+
+  /**
+   * Status code
+   */
+  status: number;
```
ghillert: rename `status` to `httpStatusCode` to minimize confusion

ui/src/app/jobs/job-execution-details/job-execution-details.component.ts
```diff
@@ -29,13 +32,19 @@ export class JobExecutionDetailsComponent implements OnInit {
    * Constructor
    *
    * @param {JobsService} jobsService
+   * @param {NotificationService} notificationService
    * @param {ActivatedRoute} route
+   * @param {LoggerService} loggerService
+   * @param {RoutingStateService} routingStateService
    * @param {Router} router
```
ghillert: Maybe add the spaces to the curly braces

ui/src/app/shared/model/error.model.ts
```diff
@@ -0,0 +1,91 @@
...
+  toString() {
+    return this.message;
+  }
+
+}
+
+/**
+ * API Error
+ */
+export class AppErrorApi extends AppError {
```
ghillert: Misleading name IMO - Could we rename it to `HttpAppError` or `HttpAwareAppError?`

ghillert:
Fixed 2 lint errors during merge - commit: dafa1fcfacd19fad4e1862095caf5230f5cd42fd

PR #771, opened 2018-05-04
Register Applications: update validation rule name

### LINKED ISSUE #770: Incorrect placeholder in app "Name" field
The placeholder in the "Name" field of the "Register Applications" form is `scdf-sample-app`. This is incorrect since it is not possible to specify a name with a "-" in it:

scdf

Here is the error when I use the same app name as the placeholder:

scdf2

oodamien:
Resolves #770


ghillert:
Verifies and merged

Issue #780, opened 2018-05-09
Stream Deploy: Sensitive properties are passed with masked values


ilayaperumalg:
When deploying the stream, the sensitive properties are passed with the masked value `*****` as deployment properties.

oodamien:
Statement:
The server return `*****` for sensitive properties.
The UI doesn’t know the real properties and have `*****` as the value set for the sensitive property.

PR #781, opened 2018-05-11
Stream deploy: Fix property issues

oodamien:
- Ignore sensitive property value on deploy stream
- Fix parsing line: app.foo=#jsonPath(payload,'$.lang')=='en' is now correctly parse
- Title dialog app property: remove version if skipper is not enable

Resolves spring-cloud/spring-cloud-dataflow-ui#780


ilayaperumalg:
LGTM, merging.

PR #787, opened 2018-05-22
Stream Deploy adjustments

### LINKED ISSUE #786: Avoid resubmitting form values when nothing changed in the stream definition
As a developer, I'd like to add support to prevent resubmitting the form-values when nothing changed in the stream definition.

This is the first-defense to add before it gets to the backend via spring-cloud/spring-cloud-dataflow#2224.

### LINKED ISSUE #788: Editing stream on skipper-mode should invoke "stream-update"
As a developer, I'd like to invoke stream-update endpoint when editing a stream on Skipper mode. For classic-mode, we must continue to undeploy and deploy the stream.

**Acceptance:**
- Decide stream-update vs. undeploy/deploy based on Skipper enabled flag.
- A platform selection is required in Skipper mode.

oodamien:
Resolves #786 :  avoid resubmitting apps form values
Resolves #788 : on stream deploy, if skipper is enabled and the stream is deployed, an update of the stream is performed instead of an undeploy/deploy.

### DIFF
--- a/ui/src/app/streams/stream-deploy/stream-deploy.component.ts
+++ b/ui/src/app/streams/stream-deploy/stream-deploy.component.ts
@@ -61,7 +61,7 @@ export class StreamDeployComponent implements OnInit, OnDestroy {
   /**
    * Original properties Array
    */
-  originalProperties: Array<string> = [];
+  ignoreProperties: Array<string> = [];
 
   /**
    * Constructor
@@ -101,7 +101,7 @@ export class StreamDeployComponent implements OnInit, OnDestroy {
         val => this.streamsService.getDeploymentInfo(val.id),
         (config: any, deploymentInfo: StreamDefinition) => {
           const properties = [];
-          const originalProperties = [];
+          const ignoreProperties = [];
 
           // Deployer properties
           Object.keys(deploymentInfo.deploymentProperties).map(app => {
@@ -134,12 +134,14 @@ export class StreamDeployComponent implements OnInit, OnDestroy {
                   keyShort = key.substring(`${appType}.`.length, key.length);
                 }
                 properties.push(`app.${app}.${keyShort}=${value}`);
-                originalProperties.push(`app.${app}.${keyShort}=${value}`);
+                ignoreProperties.push(`app.${app}.${keyShort}=${value}`);
               });
             }
           });
           this.properties = properties;
-          this.originalProperties = originalProperties;
+          if (!config.skipper) {
+            this.ignoreProperties = ignoreProperties;
+          }
           config.streamDefinition = deploymentInfo;
           return config;
         }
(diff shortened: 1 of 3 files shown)


ui/src/app/streams/stream-deploy/stream-deploy.component.ts
```diff
@@ -181,15 +188,17 @@ export class StreamDeployComponent implements OnInit, OnDestroy {
...
     const propertiesMap = {};
     value.forEach((val) => {
-      const arr = val.split(/=(.*)/);
-      if (arr.length !== 3) {
-        console.error('Split line property', val);
-      } else {
-        // Workaround sensitive property: ignored property
-        if (arr[1] === `'******'`) {
-          console.log(`Sensitive property ${arr[0]} is ignored`);
+      if (this.originalProperties.indexOf(val) === -1) {
```
ghillert: This seems to be the reason why it is not working, yet. While debugging this I saw this:

image

ghillert:
Looks like this is not working, yet. I created the following stream `t3`:

`http --server.port=9900 | splitter --expression=payload.split(' ') | log`

Went to the deployment page `http://localhost:4200/#/streams/definitions/t3/deploy`

image

Made no changes. Only pressed `Deploy the Stream`.  Does not work for freetext option either.

Looks like we create this strange property `app.splitter.splitter.expression` (see duplication of `splitter`)

ghillert:
@oodamien Something is strange. I created a stream:

image

We don't handle the use-case where no value changed. In that case the `update` REST endpoint will send back an error: `Package to upgrade has no difference than existing deployed/deleted package. Not upgrading`. Therefore, we should only activate the `Deploy the Stream` button

image

only, if anything has changed.

ghillert:
The shell does not resend the following properties:

* deployer.log.count=1
* version.http=1.3.1.RELEASE
* version.log=1.3.1.RELEASE

unless you specify them.

ghillert:
It looks like for the time being, we will always submit the properties in Skipper mode. For following stream updates we would always need to submit the properties because we would not know, if the value shown was the initial value (version `1`) or not. We would need to compare the stream to the initial version of the Stream in order to determine whether a property had changed or not.

With that the PR looks good for the changes it does add (using the Stream update REST endpoint).

ghillert:
Created https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/789 as a follow up.

ghillert:
Merged with commit 161c4e5151a36be687e1602f898a93887b2ca81f

ilayaperumalg:
This needs to be back ported to 1.5.x branch.

ilayaperumalg:
Back ported the changes 2ea34a6efa4c8a60b62118a44ef021286dc68500

PR #792, opened 2018-05-25
Stream deploy: skipper - update stream do not submit a property without change

### LINKED ISSUE #786: Avoid resubmitting form values when nothing changed in the stream definition

oodamien:
Resolves #786


ghillert:
Rebased and Merged with commit 0c42d39abce5d92606113683745ea3502bfee867 backported to `1.5.x`

PR #799, opened 2018-06-02
Better graph validation and auto link adding on transition editing

aclement:
This PR does three things:

1) Upgrade Flo - this pulls in a version that will continue to drive validation even if the graph-to-text conversion fails. This is helpful because often the validation errors will tell the user what they need to fix to make the graph-to-text conversion work!
How to test: Paste this into the DSL textbox: `<a:timestamp || b:timestamp>`. Now delete the link from `b` to `$END`. Without this PR there is no indication on node `b` that there is a problem. With this change you will get the validation error `Must have an outgoing link` on `b`.

2) Auto create a missing default link when feasible - if you adjust the exit status code for a link from a node such that there is no longer a default (non-exit status code) link from that node then we will try to add a default link automatically. We will only do this if we can come up with a suitable end target for that link. It doesn't mean just creating a link to `$END` because you could be working anywhere in the graph, what we will do right now is if the target node for the link you are adjusting has a default link from it, we will treat that as the new target for a default link we create from the original node that no longer has a default link.
How to test: Paste this into the DSL textbox: `<extractDB: timestamp || extractFiles: timestamp> && merge: timestamp && cleanse: timestamp && <writetohdfs: timestamp || transform: timestamp && writetos3: timestamp 1->foo:timestamp && bar: timestamp>` now edit the properties of the link from `writetos3` to `bar`, change exit status condition to `2`.  You will see the link you edited becomes dotted and a new default link is created from `writetos3` to `$END` (because the target of `bar` was `$END`).  Without this change you will be missing that default link and you will have no idea you need to add it... and so onto (3)

3) An extra validation message - this will put an error on a node if there is no 'default' link from that node. A default link is one with no exit status condition specified (so a solid black link).
How to test: Paste this into the DSL textbox: `<extractDB: timestamp || extractFiles: timestamp> && merge: timestamp && cleanse: timestamp && <writetohdfs: timestamp || transform: timestamp && writetos3: timestamp 1->foo:timestamp && bar: timestamp>`.  Now delete the link from `bar` to `$END`. Now edit the link between `writetos3` and `bar` and specify an exit status. When you close that properties view for the link we can't come up with a good target to create a default link for you, so we don't and you will see the validation message on `writetos3` that says `Must have at least one outgoing link with no Exit Status condition specified`.

Resolves spring-cloud/spring-cloud-dataflow-ui#804

### DIFF
--- a/ui/src/app/tasks/components/flo/render.service.ts
+++ b/ui/src/app/tasks/components/flo/render.service.ts
@@ -250,12 +250,68 @@ export class RenderService implements Flo.Renderer {
 
     if (element instanceof joint.dia.Link && element.attr('metadata')) {
       if (changedPropertyPath === 'props/ExitStatus') {
+        // If the exitstatus has been changed from blank to something then this
+        // may leave no default link from the node at the source of the link since 'element' was
+        // previously the default link. In this case add a new default link if a suitable
+        // target can be determined. If no target can be computed, a validation error will remind
+        // the user that they must add one.
+        var newExitStatus = element.attr('props/ExitStatus') || '';
+        var currentLabels = element.get('labels');
+        var currentText = '';
+        try {
+            currentText = currentLabels[0].attrs.text.text;
+        } catch (e) {
+            // Label or label value not accessible (so not set)
+        }
+        if (newExitStatus.length !== 0 && currentText.length === 0) {
+            var hasDefaultLink = false;
+            const relatedLinks = paper.model.getConnectedLinks(element.get('source'),{outbound: true});
+            for (let i = 0; i < relatedLinks.length; i++) {
+                const relatedLink = relatedLinks[i];
+                if (relatedLink === element) continue;
+                const exitStatus = relatedLink.attr('props/ExitStatus') || '';
+                if (exitStatus.length === 0) {
+                    // This is a 'default' link
+                    hasDefaultLink = true;
+                    break;
+                }
+            } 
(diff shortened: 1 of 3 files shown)


aclement:
@sabbyanandan you might find the how-to-test steps useful for knowing how things should work now.

ui/src/app/tasks/components/flo/render.service.ts
```diff
@@ -250,12 +250,68 @@ export class RenderService implements Flo.Renderer {
...
+                    hasDefaultLink = true;
+                    break;
+                }
+            } 
+            if (!hasDefaultLink) {
+                // Create a new default (no specified exit status) link 
+                const newDefaultLink = this.createLink();
+                const sourceNodeId = element.get('source').id;
+                const targetNodeId = element.get('target').id;
+                const targetModel = paper.getModelById(targetNodeId);
```
BoykoAlex: `targetModel` is unused. Will remove before merging

BoykoAlex:
@sabbyanandan shall I merge this one or you wanted to have a look at it as well before it's merged?

sabbyanandan:
My usual tests look good and the reported problem is resolved now. When I also explicitly set multiple transitions without any default, I get alerted about it, too. 

image

Thanks, @aclement. @BoykoAlex: please merge it at your convenience.

BoykoAlex:
Merged with 90908f445443e59ffdf66e0b3061b8138903d18d

PR #805, opened 2018-06-05
Refactoring About Module

### LINKED ISSUE #801: Bring About tab to consistent UI/UX

oodamien:
Resolves #801


ui/src/app/about/components/about-more/about-details.component.html
```diff
@@ -0,0 +1,261 @@
...
+          <span>Spi Version:</span>
+          <strong>{{ dataflowVersionInfo.runtimeEnvironment.appDeployer.deployerSpiVersion }}</strong>
+        </div>
+        <div class="line">
+          <span>Java Version:</span>
+          <strong>{{ dataflowVersionInfo.runtimeEnvironment.appDeployer.javaVersion }}</strong>
+        </div>
+        <div class="line">
+          <span>Platform Api Version:</span>
+          <strong>{{ dataflowVersionInfo.runtimeEnvironment.appDeployer.platformApiVersion }}</strong>
```
ghillert: Since we don't use a table anymore, we should visually point out that no value is available. So may a `N/A`?

PR #815, opened 2018-06-12
Adopt JointJS 2.x for Flo

BoykoAlex:
Fixes https://github.com/spring-projects/spring-flo/issues/30
Needs to be adopted on SCDF UI side

@sabbyanandan there is no issue on the SCDF UI side for this.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -48,7 +48,7 @@
     "web-animations-js": "2.3.1",
     "stompjs": "2.3.3",
     "jshint": "2.9.5",
-    "spring-flo": "git://github.com/spring-projects/spring-flo#483faf1a1b6ed595580a2af86d901d3f33013353",
+    "spring-flo": "git://github.com/spring-projects/spring-flo#9f3c37be645a96d6073bc42299cefa6afd52d7a4",
     "ng-busy": "1.4.4"
   },
   "devDependencies": {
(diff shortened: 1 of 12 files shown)


oodamien:
1. Create the following stream: ok
`time --trigger.time-unit=SECONDS --spring.cloud.stream.bindings.applicationMetrics.destination=metrics | file --file.name=test.txt --file.directory=/Users/PATH_USER/Desktop --spring.cloud.stream.bindings.applicationMetrics.destination=metrics`

2. Deploy the stream: ok

3. Checking test.txt: ok

4. Display the metrics in the Flo Chart: ko
image

The counters are not presents.

sabbyanandan:
@oodamien: Do you have [Metrics Collector](https://github.com/spring-cloud/spring-cloud-dataflow-metrics-collector#e2e-cheat-sheet) running and is reachable to SCDF via `localhost:8080` (unless overridden)?

sabbyanandan:
It is a bit odd it shows in gray instead of a green border, though. The gray scale shows it is either not deployed, partial, or failed.

BoykoAlex:
@oodamien I think you need t reload the page to see the message rates etc. I believe the screenshot is taken when stream status is deploying. There is a bug in streams list page that its contents are not refreshed periodically like others... Hence you need to reload it manually and then when stream status is deployed you'd be able to see the message rates, proper colours etc.

PR #822, opened 2018-06-19
Stream Creation: deploy option conditional show

### LINKED ISSUE #783: Shortcut stream deployment should hide "deploy" option if >1 backend is available
As a user, while deploying stream from Flo-dialog with multiple deployment backends, I'd like to have the option to select the backend. Today, if I deploy it from the Flo-dialog, the first in the list is used by default, but that won't be the intended platform always.

A possible solution would be to include deployment backends in the drop-down if >1 is available. _(For a single backend, we use the single-entry by default anyway)_

image
image

oodamien:
The deploy option is not available if skipper is enabled and more than one platform available
Resolves #783

### DIFF
--- a/ui/src/app/streams/stream-create/create-dialog/create-dialog.component.html
+++ b/ui/src/app/streams/stream-create/create-dialog/create-dialog.component.html
@@ -49,7 +49,7 @@
           </div>
         </div>
       </div>
-      <div class="row row-stream-deploy">
+      <div class="row row-stream-deploy" *ngIf="isDeployEnabled">
         <div class="col-sm-offset-4 col-sm-18">
           <label class="checkbox-inline">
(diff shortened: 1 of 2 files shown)


BoykoAlex:
It got assigned to me for review and I for some reason decided it's assigned to me to work on it... sigh...

Changes look good to me :-)

I think we should probably see if we can take the tests i created while working on this by mistake...

I'll push the test spec separately once this is merged.

BoykoAlex:
I've been assigned to review this, but thought I'm assigned to work on it... sigh...
Changes look good to me!-)
I'll add my test spec i had while working on this by mistake in this PR before merging

BoykoAlex:
Merged with 5b0ea14b5944c4f671ec16613255a9b6457112eb

PR #839, opened 2018-06-28
Register App: URI regex fix

### LINKED ISSUE #838: Register App with a maven URL in chrome freeze the tab browser

oodamien:
Resolves #838: simplify validation URI


ghillert:
@oodamien @sabbyanandan Still think we should have slightly better validation eventually. But for now I think this is better. Merging.

PR #844, opened 2018-07-04
Move `unresolved` in ExtraMetadata object from ElementMetadata

BoykoAlex:
The `unresolved` property was set on the ElementMetadata object. It's moved inside ExtraMetadata object which is contained by ElementMetadata to ensure TS `readonly` is not violated.
New TS compiler with Angular 6 seem to catch these violations...


ghillert:
Rebased and merged with commit 11b4a42cfc85d611ea3c176614355bfb8fe65344

Issue #795, opened 2018-05-31
Upgrade to Angular 6


ghillert:
Should be fairly straight-forward. This also include an upgrade of RxJS 6. Hopefully this will give us a smaller prod bundle size.

See also: https://blog.angular.io/version-6-of-angular-now-available-cc56b0efa7a4

sabbyanandan:
Depends on spring-projects/spring-flo#31

PR #846, opened 2018-07-06
gh-795 Migrate to Angular 6

ghillert:
* Convert configuration files from `.angular-cli.json` to `angular.json`
* Move E2E tests to `e2e/src`
* Move Protractor configuration files to `e2e` folder
* Refactor error handling classes due to stricter type-checking of generics
* Add `rxjs-compat` to avoid having to refactor RxJs to use pipeable operators

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/795


BoykoAlex:
@ghillert can you try this branch of Flo please: https://github.com/spring-projects/spring-flo/tree/angular6-compiler

It includes previous changes and also has a complete rework of building Flo lib for consumption. I'm finally using `ngc` from Angular 6 instead of 4.4.6. The Flo bundle must be smaller now since i trimmed off some `codemirror` dependencies that sneaked into the bundle file. I have built SCDF UI with the latest Angular 6 compiled Flo and the size is 3.5Mb which i think is a great sign :-)

ghillert:
@oodamien Created follow up issue https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/847

Issue #852, opened 2018-07-12
Use a dedicated UUID/GUID library


ghillert:
We should use a dedicated UUID/GUID library, something like https://github.com/kelektiv/node-uuid 2kb. I don't like maintaining that type of code. This should get rid of the lint errors as well:

```
Forbidden bitwise operation
```

PR #853, opened 2018-07-12
gh-852 Use a dedicated UUID/GUID library

ghillert:
* Add `node-uuid` dependency
* Update code
* Ensure that lint errors pertaining to `Forbidden bitwise operation` no longer occur

resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/852

### DIFF
--- a/ui/src/app/shared/services/group-route.service.ts
+++ b/ui/src/app/shared/services/group-route.service.ts
@@ -1,8 +1,10 @@
 import { Injectable } from '@angular/core';
 import { LocalStorageService } from 'angular-2-local-storage';
 
+import * as uuidv4 from 'uuid/v4';
+
 /**
- * A service for global logs.
+ * A service for create group route.
  *
  * @author Damien Vitrac
  */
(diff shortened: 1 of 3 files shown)


ui/src/app/shared/services/group-route.service.ts
```diff
@@ -13,10 +15,7 @@ export class GroupRouteService {
   }
 
   create(args): string {
-    const key = `group-${'xxxxx-xxxxx-xxxxx-xxxxx'.replace(/[xy]/g, function (c) {
-      const r = Math.random() * 16 | 0, v = c === 'x' ? r : (r & 0x3 | 0x8);
-      return v.toString(16);
-    })}`;
+    const key = uuidv4();
```
oodamien: The UUID should respect the following format: `xxxxx-xxxxx-xxxxx-xxxxx`
The `isSimilar` method is used to define if the GET params is a stream definition or a group (UUID).

Issue #836, opened 2018-06-27
Importing maven resourse, where the group id starts with "n"


kalvarez2:
Hi
If you try to use the bulk import apps, from properties, and your app maven identifier starts with n, the application does not get registered correctly.
For instance, trying to bulk import an application with the following descriptor will fail:
source.personSource=maven://net.sample:person-source:0.0.1-snapshot

I believe the bug can be fixed by changing:
https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/master/ui/src/app/apps/apps-bulk-import/properties/apps-bulk-import-properties.component.ts
line 91, to split by \n, and not by /n.

I think the intention was to break on new line, which this would accomplish.
thanks

kalvarez2:
Hi, 
Any update on this? Should I post it elsewhere? 
thanks!

sabbyanandan:
@kalvarez2: Apologies for the delayed response. Thanks for reporting it, though. Feel free to contribute a PR with the suggestion, we would be happy to review and take it from there.

PR #863, opened 2018-07-18
Fix for Issue #836, bulk import apps

kalvarez2:
Fix for bulk import apps when the artifact identifier starts with N


oodamien:
Thanks @kalvarez2

kalvarez2:
np, glad to be of service

PR #774, opened 2018-05-07
UI: left navigation

oodamien:
SPIKE global UI/UX

Navigation:
- Update the main navigation from the top to the left
- 2 states: small (icons) / full (icons + texts)
- Update logo Data Flow.
- Update footer to have a full screen approach.
- Add quick search (search in apps/streams/tasks)

Forms (App Forms):
- Display error after submit the form
- The submit button is alway enabled. If there is error in the form, a message is displayed

App Register/Bulk Import:
- Revisit the way we add applications: one entry point
- Review URL navigation

Stream / Task Create:
- Width content: fullscreen stream create (100% width & height), avoid double scroll
- Polish editor UI

List:
- Action row: dropdown

Search:
- Revisit UI/UX
- Dedicate component

Layout:
- Introduce 3 layout: medium / large / full

resolves spring-cloud/spring-cloud-dataflow-ui#873
resolves spring-cloud/spring-cloud-dataflow-ui#226
resolves spring-cloud/spring-cloud-dataflow-ui#823
resolves spring-cloud/spring-cloud-dataflow-ui#225
resolves spring-cloud/spring-cloud-dataflow-ui#617

PR #881, opened 2018-08-29
App Details: add all properties

### LINKED ISSUE #879: Add an extra step in the UI to confirm whether or not to show all Boot properties
As a user, if I don't register the App with companion artifact, I'd like to be warned about the available Boot properties as possibilities. 

Currently, in the Shell, if we don't provide the companion artifact for an App, we show the following message on autocomplete trigger. It'd be good to adopt a similar pattern in the UI, so users can confirm whether or not they'd like to see all properties. 

> dataflow:>stream create foo --definition "http | userprocessor -- (HIT-TAB-BUTTON)
> dataflow:>stream create foo --definition "http | userprocessor --Display all 1123 possibilities? (y or n)

oodamien:
Resolves #879 
Wait for merge #774

### DIFF
--- a/ui/src/app/apps/app-details/app-details.component.html
+++ b/ui/src/app/apps/app-details/app-details.component.html
@@ -58,40 +58,53 @@
           <strong>Version</strong>: {{ detailedAppRegistration.version }}
         </div>
       </div>
-      <div *ngIf="detailedAppRegistration?.options && detailedAppRegistration?.options.length > 0">
-        <table id="table-properties" class="table table-hover" *ngIf="detailedAppRegistration?.options">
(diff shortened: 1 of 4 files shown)


tzolov:
@oodamien i've compiled the SCDF-UI with the PR and added it as dependency to the SCDF local server. But when i try to start i get this (see the attachment). Am  I forgetting something? 
screen shot 2018-08-30 at 7 39 56 pm

tzolov:
LGTM, 
I've performed manual tests. The property threshold rules is applied as expected in app-info, stream-create-settings and stream-deploy-app-properties contexts

PR #892, opened 2018-09-10
Jobs Service / Tasks Service refactoring

### LINKED ISSUE #796: Move JSON deserialization for Job-related calls to domain model
Since we do it already for several domain model classes, we should refactor the deserialization of Job/Task related REST calls to the respective domain model classes, instead of doing it in the services.

This allows for better isolation and easier testing. Therefore, as part of this issue we should also add unit tests.

oodamien:
Moved parser code from the API services to the models
Added unit tests for Job/Task models
Resolves #796


ghillert:
LGTM - merged with commit 9e7997008ca0569c28f80a3e0b81a2a9a18eca1a

PR #904, opened 2018-09-17
Audit Records: truncate data column

### LINKED ISSUE #903: Wrap overflowing text on the audit list view
As a user, while verifying the audit trails for a stream with deployment properties, I noticed a horizontal scroll-bar, which is not a very good UX.

image

oodamien:
Resolves #903


ghillert:
LGTM - merging

PR #924, opened 2018-10-01
Tasks: update message and visibility creation action

### LINKED ISSUE #921: Unable to create new Tasks
As a user, while attempting to create Tasks, I notice that the "Create Task" button is hidden (see below), and it won't show up until Task Apps are registered. 

I don't get notified about it, though. I have had to assume and register Tasks to unhide the button. However, I do see a message when there isn't any Task definitions.

image

oodamien:
Resolves #921


ghillert:
LGTM - created a follow-up issue though for the server-side right now: https://github.com/spring-cloud/spring-cloud-dataflow/issues/2463 Might also lead to a UI issue.

Merged with commit: 4a3fe1190044020717ad5441d44712946f4601b5

PR #948, opened 2018-10-11
Upgrade Angular to 6.1.10

### LINKED ISSUE #923: Update to Angular 6.1.9
https://blog.angular.io/angular-v6-1-now-available-typescript-2-9-scroll-positioning-and-more-9f1c03007bb6

ilayaperumalg:
Resolves #923


ilayaperumalg:
Tried to upgrade the `rxjs` module to `6.3.3` and it seems more involved and hence hasn't updated it.

Also, based on `https://update.angular.io/`, it is suggested to remove deprecated RxJS 6 features. But, that throws a large number errors and I guess we need to do it at the beginning of the 1.8.x.

PR #950, opened 2018-10-16
Task / Scheduler: enforce rules for deployment properties

### LINKED ISSUE #937: Scheduling - Enforce rules for deployment properties
Currently you can enter properties like the following and you can press "Schedule the task". This leads to a `500` error:

image

The error is:

```
[{
    "logref": "IllegalArgumentException",
    "message": "Only deployment property keys starting with 'app.', 'deployer.' or, 'scheduler.' allowed, got 'prop2'"
}]
```

oodamien:
Resolves #937
Deployment properties: the key should start with ‘app.’ or ‘deployer.’ or ‘scheduler.’

### DIFF
--- a/ui/src/app/tasks/task-launch/task-launch.component.html
+++ b/ui/src/app/tasks/task-launch/task-launch.component.html
@@ -31,8 +31,8 @@
                   <td>
                     <div [class.has-error]="form.get('args.' + i).invalid">
                       <input formControlName="key" type="text" class="form-control"/>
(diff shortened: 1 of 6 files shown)


jvalkeal:
Looks ok. Originally I thought I'd nitpick because checks is really naive as passing just`app.` would be rather dump thing which check allows. But as task seem to actually launch with just passing `app.`, don't feel like nitpicking about it. Merging...

PR #955, opened 2018-10-18
Stream create button disabled. Flo with codemirror-minified

### LINKED ISSUE #951: Non-intuitive error message when duplicate apps are in the streaming DSL
While creating a pipeline made of duplicate "streaming apps", I encountered an error message, but it is not clear as to what is wrong.

image

BoykoAlex:
Fixes #951 
1. Error must be displayed either in DSL or on the graph in the built version. Without this fix `npm start` (local UI server) works fine, but the built version doesn't work on port 9393
2. The `Create Stream(s)` button should be disabled if invalid stream definitions are present. Clicking on the disabled button brings up the tooltip with the corresponding message


oodamien:
The tooltip is displayed even if there is no error.
The submit button should not be disabled in all case, if there is error (invalid stream definitions), an error message (toast) should be display.

BoykoAlex:
@oodamien I've just tried it and seems to be working as I'd expect it too... Feels like we are looking at different buttons perhaps.
I've been looking at this one:
screen shot 2018-10-19 at 09 37 25

BoykoAlex:
BTW, I forgot to put Flo bump to 0.8.5. Without that upgrade it is impossible to see when the graph has errors and hence it may seem like `Create Stream` button is disabled if there are no errors.
If we'd rather display a toast error message than disabling the button we can just scrap this PR :-) I saw that Gunnar has Flo bumped to 0.8.5 in a different PR and that version upgrade address the bigger issue of error not displayed in the stream editor.

PR #974, opened 2018-11-05
Clean RxJS usages

### LINKED ISSUE #538: Update rxjs operator imports to use lettable operators
After migrating to Angular 5 we need to update the rxjs operators imports to the new format.  For example:
```
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/operator/map';
import 'rxjs/add/operator/filter';
```
can be replaced with 
```
import { Observable } from 'rxjs/Observable';
import { map, filter } from 'rxjs/operators';
```

oodamien:
Use lettable operators
Resolves #538


ghillert:
LGTM - merged

PR #999, opened 2018-12-03
Upgrade Flo to 0.8.6 and get the fix for overlapping links

### LINKED ISSUE #996: Connectors bend and overlap for duplicate streams that start with named-destinations
As a user, while copying the same stream with named-destinations a few times, I see the connectors bend, overlap with one another, and are basically messed up.

image

**Should we instead error saying duplicate streams?**

When I change the destination to something else, they automatically get adjusted.

image

BoykoAlex:
Fixes #996 
Also upgrade Flo to angular 7.1.1 and JointJS 2.2.1

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -49,7 +49,7 @@
     "web-animations-js": "2.3.1",
     "stompjs": "2.3.3",
     "jshint": "2.9.5",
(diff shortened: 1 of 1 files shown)


oodamien:
LGTM, merged

PR #1010, opened 2018-12-20
Remove deploy checkbox from create stream dialog

### LINKED ISSUE #1009: Short-cut to "deploy" from Flo-canvas is enabled on Skipper mode
As per https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/783, we wanted to hide the "Deploy" button when the backend is on Skipper mode _and_ there are >1 platforms defined as part of the deployment options; however, on 2.0 (`master`), there's no classic-mode (skipper-mode is the default), and with that I see the "Deploy" button is visible.

This would confuse when >1 platform is defined, of course. What would be the default platform the stream(s) would be deployed against? 

I guess that the decision-flag to decide whether the backend is in skipper vs. classic mode has been removed (in 2.0), and we would have to find a different approach now.

image

BoykoAlex:
Fixes #1009 

Remove deploy checkbox and its backend support and tests

### DIFF
--- a/ui/src/app/streams/stream-create/create-dialog/create-dialog.component.html
+++ b/ui/src/app/streams/stream-create/create-dialog/create-dialog.component.html
@@ -50,26 +50,15 @@
           </div>
         </div>
       </div>
-      <div class="row row-stream-deploy" *ngIf="isDeployEnabled">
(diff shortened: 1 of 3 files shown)


oodamien:
LGTM

PR #1014, opened 2018-12-21
Navigation: loading state

### LINKED ISSUE #1013: Minor UX improvement: Add loading indicator for left navigation
With security and skipper enabled, it may take a second or so for the left navigation to appear, leaving only the `Apps` menu item. Eventually the items appear, but the gap is long (1-2sec) enough that I assume something went wrong.

Maybe we may consider adding loading indicators for asynchronous components.

image

oodamien:
Resolves #1013

### DIFF
--- a/ui/src/app/layout/navigation/navigation.component.html
+++ b/ui/src/app/layout/navigation/navigation.component.html
@@ -1,68 +1,79 @@
-<app-navigation-search [force]="!isCollapsed" [dataflowAppRoles]="['ROLE_VIEW']"></app-navigation-search>
+<div *ngIf="featureInfo$ | async as featureInfo; else loading">
+  <app-navigation-search [force]="!isCollapsed" [dataflowAppRoles]="['ROLE_VIEW']"></app-navigation-search>
 
-<div class="sidebar-user" *ngIf="securityInfo.isAuthenticationEnabled" [dataflowAppRoles]="['ROLE_VIEW']">
-  <div dropdown class="sidebar-item dropdown" *ngIf="securityInfo.isAuthenticated">
-    <a class="dropdown-toggle" dropdownToggle>
-      <i class="sidebar-fa fa fa-user"></i>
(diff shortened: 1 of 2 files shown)


ui/src/app/layout/navigation/navigation.component.html
```diff
@@ -1,74 +1,85 @@
...
 
-<div class="sidebar-user" *ngIf="securityInfo.isAuthenticationEnabled" [dataflowAppRoles]="['ROLE_VIEW']">
-  <div class="sidebar-item" routerLinkActive="active" *ngIf="!securityInfo.isAuthenticated">
-    <a routerLink="login">
-      <i class="sidebar-fa fa fa-sign-in"></i>
-      <span class="text">Login</span>
-    </a>
+  <div class="sidebar-user" *ngIf="securityInfo.isAuthenticationEnabled" [dataflowAppRoles]="['ROLE_VIEW']">
+    <div class="sidebar-item" routerLinkActive="active" *ngIf="!securityInfo.isAuthenticated">
+      <a routerLink="login">
```
ghillert: I don't think we need that anymore. As soon as you try to access the UI and you are NOT authenticated, you will be redirected to the UAA server.

ghillert:
Please rebase (There is a conflict).

Also, I think the problem still persists:

Set a breakpoint in `org.springframework.cloud.dataflow.server.controller.AboutController` to simulate a slow connection.

```java
 @RequestMapping(method = RequestMethod.GET)
	@ResponseStatus(HttpStatus.OK)
	public AboutResource getAboutResource() {
...
```

You will observe this:

image

Not sure if this preventable entirely, or if we should have some sort of visual loading indicator.

ghillert:
@oodamien Can we apply the same logic as you did with https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/1107 That would solve the issue nicely.

PR #1018, opened 2019-01-03
Adjust client side parser for unbound app handling

### LINKED ISSUE #1005: Switching from comma to double-pipe in streaming-app-dsl doesn't work
Building on the foundations of https://github.com/spring-cloud/spring-cloud-dataflow/pull/2629 (via CF-server v.1.7.3), the Dashboard doesn't accept `||` instead of `,` in the textual-DSL section. It appears the UI is validating explicitly on the `,` instead of relying on the backend. 

Maybe it is intentionally done on the client-side, but it is out of sync with the backend changes in v1.7.3.

cc:/ @aclement

aclement:
This change adjusts the client side syntax to '||' from
',' for unbound apps so that it matches the server side.
It has adjusted the tokenizer, the parser as well as the
text-to-graph and graph-to-text code. Tests have been updated
to recognize the new syntax and messages.

Fixes #1005

### DIFF
--- a/ui/src/app/shared/services/parser.spec.ts
+++ b/ui/src/app/shared/services/parser.spec.ts
@@ -102,7 +102,7 @@ describe('parser:', () => {
     });
 
     it('long running apps', () => {
-      parseResult = Parser.parse('aaa, bbb', 'stream');
+      parseResult = Parser.parse('aaa|| bbb', 'stream');
       expect(parseResult.lines.length).toEqual(1);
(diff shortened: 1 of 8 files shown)


BoykoAlex:
Merged with:
7e228a1794f7caf0cd9702fa48444150096e42ae for 2.x
5fcb743a61a11238cf96880cb9f0cbf6196586e1 for 1.7.x

PR #1030, opened 2019-01-17
Bring stream parser client side in line with server side

aclement:
Per issue 2499 the server side handling of special chars in
argument values changes. This is the client side change that
brings the client side parser in line with that.

Resolves #2499


BoykoAlex:
Merged with 5a7421bae070857971f6567cb8fab9e5d5e17f1a

PR #1041, opened 2019-01-25
Grafana Integration

### LINKED ISSUE #1027: Add a link to launch grafana dashboard for each stream
As a developer, I'd like to explore options to investigate whether if we can embed Grafana dashboard widgets in the Dashboard. 

Notes:
- This requires to be validated on top of spring-cloud/spring-cloud-dataflow#2753.
- For this to work, we may have to come up with the dynamic Grafana URL based on the selected stream for context. That'd mean, we would have to formulate the URL with stream-specific unique identifiers and query properties, so when Grafana loads, it will automatically also query and render the widgets automatically for the selected stream.
- We would also want to come up with the ideas to query the `/metrics` endpoint directly to continue to render the message rates in the UI natively (depends on spring-cloud/spring-cloud-dataflow#2737 and spring-cloud/spring-cloud-dataflow#2736).

Depends On https://github.com/spring-cloud/spring-cloud-dataflow/issues/2786

oodamien:
Created Grafana Module (Shared)
Updated Streams / Stream pages
Updated FeaturedInfo, AboutInfo models
Resolves #1027 
Depends on https://github.com/spring-cloud/spring-cloud-dataflow/pull/2787


tzolov:
Rebased, squared and merged at 8072c640a0d92b176cf847eb61d62e62c09d90b9

PR #1060, opened 2019-02-04
Unregister all the applications

### LINKED ISSUE #1049: Revise delete-all operation to rely on the new endpoint
As a developer, I'd like to revise the delete-all operation from the Dashboard to use the endpoint defined via spring-cloud/spring-cloud-dataflow#2507, so both the Shell/GUI offers a consistent UX.

Notes:
- Use the new endpoint as opposed to deleting one app at a time, thus eliminating "n" backend calls. It should improve performance, and likewise, we won't run into DB connection limit issues.
- Rely on the REST response to highlight the outcome of the backend operation.
- Remove any client-side overhead.

oodamien:
Added unregister all
Updated tests
Resolves #1049

### DIFF
--- a/ui/src/app/apps/apps-unregister-all/apps-unregister-all.component.html
+++ b/ui/src/app/apps/apps-unregister-all/apps-unregister-all.component.html
@@ -1,0 +1,26 @@
+<div class="modal-content">
+  <div class="modal-header">
+    <h4 class="modal-title pull-left">Confirm Unregister all Applications</h4>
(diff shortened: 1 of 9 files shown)


ilayaperumalg:
@oodamien I am not sure if we had incorporated the recent change to `unregisterAll(withSelectedApps)`. Could you confirm?

oodamien:
As discussed, `unregister All()` doesn't have any parameter but delete all the apps.

ilayaperumalg:
That means we can't have the `unregister` as an action for a selected number of apps as it is only all or none.

PR #1078, opened 2019-02-15
Stream builder: tooltip description for deployment platform

### LINKED ISSUE #1077: Format of deployer-properties is expected differently
As a user, when I deploy a stream from the UI with deployer property overrides (as described in the docs), they don't get propagated over to the apps. The same does when we do it from the Shell.

Observations:

In the Dashboard, we have a short-form for deployer properties. For example, to bind `jdbc-sink` with a custom service-instance (lets say: `mysqlcups`), I'd pass it in the Shell as `deployer.jdbc.cloudfoundry.services=mysqlcups`; however, when passing it from the UI, the shortcut is to supply `cloudfoundry.services` as the key, and `mysqlcups` as the value for just `jdbc-sink` application. (see screenshot below)

Unfortunately, it is not documented. There are no hints of any kind in the UI either. Users could blindly paste `deployer.jdbc.cloudfoundry.services=mysqlcups` as the key and value, or try different combinations, which is bad. 

It'd be good to highlight in a hint on what is expected. Perhaps use the same example above in the hint, so it is descriptive to the users.

image

oodamien:
Resolves #1077

### DIFF
--- a/ui/src/app/streams/stream-deploy/builder/builder.component.html
+++ b/ui/src/app/streams/stream-deploy/builder/builder.component.html
@@ -69,15 +69,28 @@
 
             <!-- 5 -->
             <div class="line-toggle">
+              <div class="help">
+                <span class="fa fa-question-circle" tooltipPlacement="right" [tooltip]="tooltipTemplate"></span>
+                <ng-template #tooltipTemplate>
+                  <div>
+                    <div><strong>Example:</strong></div>
+                    <div>Key: <strong>cloudfoundry.services</strong></div>
+                    <div>Value (application <strong>jdbc</strong>): <strong>mysqlcups</strong></div>
+                    <div>Generate property: <strong>deployer.jdbc.cloudfoundry.services=mysqlcups</strong></div>
+                  </div>
+                </ng-template>
+              </div>
+
+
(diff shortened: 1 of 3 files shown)


ui/src/app/streams/stream-deploy/builder/builder.component.html
```diff
@@ -69,15 +69,28 @@
...
             <!-- 5 -->
             <div class="line-toggle">
+              <div class="help">
+                <span class="fa fa-question-circle" tooltipPlacement="right" [tooltip]="tooltipTemplate"></span>
+                <ng-template #tooltipTemplate>
+                  <div>
+                    <div><strong>Example:</strong></div>
+                    <div>Key: <strong>cloudfoundry.services</strong></div>
+                    <div>Value (application <strong>jdbc</strong>): <strong>mysqlcups</strong></div>
+                    <div>Generate property: <strong>deployer.jdbc.cloudfoundry.services=mysqlcups</strong></div>
```
ilayaperumalg: We can say `Generated` instead of `Generate`. Will fix on merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as 46fe0c83

PR #1091, opened 2019-02-26
Stream deploy: force platform selection if more than 2 platforms

### LINKED ISSUE #1087: Force platform selection when >1 option is available for stream deployments
As a user, when deploying a stream with 4 platform-backends, I'm not forced to make a selection. Instead, the deployment picked the first platform from the list, and that is not the platform I wanted to deploy.

We would have to replicate what we did in Tasks via spring-cloud/spring-cloud-dataflow-ui#1075, but for streams.

oodamien:
Resolves #1087

### DIFF
--- a/ui/src/app/streams/stream-deploy/builder/builder.component.html
+++ b/ui/src/app/streams/stream-deploy/builder/builder.component.html
@@ -192,7 +192,7 @@
                   ({{ platform.type }})
                 </option>
                 <option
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
@oodamien The platform drop-down list has an item named `invalid` which can be removed
image

oodamien:
@ilayaperumalg Thanks, I amended my last commit.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 96751ce8

PR #1093, opened 2019-02-26
Fix slection of group for app nodes

### LINKED ISSUE #1092: Stream editor confuses `counter` processor and sink
Paste the following DSL into the text area:
```
file | counter
twitterstream | counter | cassandra
```
The resultant graph is:
screen shot 2019-02-26 at 11 38 11

Although it's impossible to figure correctly `processor` or `sink` in general. Perhaps, in this specific case the `sink` should be picked because there are no more nodes after it. Only text -> graph is affected. I believe this is an isolated issue with one utility function.

BoykoAlex:
Fixes #1092


tzolov:
LGTM With the fix i can create and deploy streams the counter app. as a sink and as a processor as.

tzolov:
Rebased and merged at 0d3f23a9fbb3a0fdb7a00f760164663b39e5cfe0

PR #1101, opened 2019-03-01
Fix Lint errors

### LINKED ISSUE #1100: Fix Lint Errors

oodamien:
Resolves #1100


ilayaperumalg:
@oodamien While some of the lint fixes are direct, there are some changes that look like refactoring (at least from the first look one-to-one comparison). Can we postpone this change post GA to avoid any surprises?

ilayaperumalg:
Looks better now. Merging.

PR #1155, opened 2019-03-25
Stream creation: add custom URL params encoder

### LINKED ISSUE #1074: Stream/Task Definitions with Plus Sign Not Created Correctly
When creating a stream or task with a property that has a plus sign as a character in its value the plus sign is replaced with a space upon publishing the stream/task. I'm currently running SCDF 1.7.4 Local, and have observed this in 1.6.3 and 1.7.3. Manually editing the task/stream definition in the database fixes the issue.

### LINKED ISSUE #1152: Stream creation fails when using property value with string concatenation
When creating the following stream:

http | transform --expression='Hello '+payload.toUpperCase()+'!' | log

the server fails the stream definition with the following error:

```
ataflow-server    | 2019-03-22 06:54:03.040  WARN 1 --- [nio-9393-exec-8] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request: 115E:(pos 58): unexpected data in stream definition '('
dataflow-server    | http | transform --expression='Hello ' payload.toUpperCase() '!' | log
dataflow-server    |                                                           ^
dataflow-server    |
```

As you can see from the above the DSL string doesn't have the String concatenation symbol `+` with it.
The same works from the SCDF shell.

oodamien:
Closes #1152 
Closes #1074

### DIFF
--- a/ui/src/app/shared/support/encoder.utils.ts
+++ b/ui/src/app/shared/support/encoder.utils.ts
@@ -1,0 +1,24 @@
+import { HttpParameterCodec } from '@angular/common/http';
+
+/**
+ * Custom HTTP Parameter Encoder
+ */
+export class DataflowEncoder implements HttpParameterCodec {
+
+  encodeKey(key: string): string {
+    return encodeURIComponent(key);
+  }
+
+  encodeValue(value: string): string {
+    return encodeURIComponent(value);
+  }
+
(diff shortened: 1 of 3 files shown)


ghillert:
LGTM - merged with commit 7cc3ed2a85b83d24a192ee9973349cff474ea592

PR #1168, opened 2019-04-04
Switch to Flo from master to fix Checkbox property control

### LINKED ISSUE #1167: Stream deploy wrongly uses boolean from freetext
`hdfs` sink uses boolean options and all works ok until I go to `freetext` and come back to `builder` when form is rebuild from raw property strings.
If I i.e. have
```
app.hdfs.file-uuid=true
```
hdfs-properties-1

I think this is caused by populate methods in `StreamDeployBuilderComponent` which are taking both keys/values as strings and then passes boolean as string. `PropertiesDialogComponent` then don't understand this. I manually tested to set property value as real boolean if text value is `true` and then `PropertiesDialogComponent` correctly shows boolean values.

This should probably get fixed somewhere else than in `StreamDeployBuilderComponent`.

BoykoAlex:
Fixes #1167

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -48,7 +48,7 @@
     "web-animations-js": "2.3.1",
     "stompjs": "2.3.3",
     "jshint": "2.9.5",
-    "spring-flo": "0.8.8",
+    "spring-flo": "https://github.com/spring-projects/spring-flo#db2d3811998eab3780cbb3a73bc363a1829f481a",
     "rxjs-compat": "6.2.1",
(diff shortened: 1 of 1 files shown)


jvalkeal:
Tried it and boolean values work ok now. I just tried this via #1169 as it moves to newer master snapshot.

jvalkeal:
Obsolete as a88a654c91729fda1831a7bf9aec87c6f2cdecb0 put flo version to a newer github one, nothing to do.

PR #1175, opened 2019-04-12
Implement deployment properties grouping

### LINKED ISSUE #1174: Implement property grouping for deployment properties
To ease how properties are show, implement properties dialog by collapsible groups.

jvalkeal:
- Add new PropertiesGroupsDialogComponent which is more
  opinionated towards showing bigger list of properties.
- Deployment properties are grouped together by group id,
  and groups are sorted to get main group first.
- Dialog will collapse first group and after that one group
  is usually kept open unless user closes already open group.
- Fixes #1174

Result is something like this:
Screenshot from 2019-04-12 08-06-15


oodamien:
Lint error:
ERROR: /ui/src/app/shared/flo/properties-groups/properties-groups-dialog.component.ts:92:53 - variable 'GroupPropertiesGroupModel' used before declaration

PR #1208, opened 2019-06-19
Stream Deploy: fix ignore properties on deploy

### LINKED ISSUE #1204: Stream deploy from dashboard uses sanitized value(s)
When a stream is configured through the shell with values that should be sanitized, and deployed from the UI, the UI is POSTing what it retrieved from the sanitizer and not allowing values to be read from the db.

In order to reproduce:
1. `stream create --name test --definition "http --mapped-request-headers='*' | parser-endpoint: httpclient --http-method=POST --headers-expression={'Content-Type':'application/json'} --url='https://parser.my.url/' | log"`
2. Deploy the stream from the UI
3. The respective httpclient app gets configured with the literal sanitized url (e.g. httpclient.url:"`******`")

oodamien:
Resolves #1204

Fix bugs:
- The app properties loaded from the DSL should not be submitted
- The sensitive value properties should not be submitted
- Error on showing the app properties modal

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,7 +101,7 @@
 							<goal>install-node-and-npm</goal>
 						</goals>
 						<configuration>
-							<nodeVersion>v10.8.0</nodeVersion>
+							<nodeVersion>v10.16.0</nodeVersion>
 						</configuration>
 					</execution>
 					<execution>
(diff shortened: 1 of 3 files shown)


jvalkeal:
lgtm, but yeah this sensitive feature is a bit of a hack. Properties dialog shows `******` but as soon as you modify it either in dialog or freetext, it stays there cleartext.

PR #1262, opened 2019-08-08
Task schedule creation: fix concatenation args

### LINKED ISSUE #1242: Supplying Task arguments with new-lines breaks the property parsing
As a user, when scheduling a Task with arguments that include line-breaks, I notice that the property parsing produces incorrect schedule behavior.

Supplying the multiple key-value pairs separated by commas like this works:

> localFilePath=/var/scdf/atrades-shared-files/ tempDir=/home/vcap/

However, passing them as separate lines fail:

> localFilePath=/var/scdf/atrades-shared-files/ 
> tempDir=/home/vcap/

oodamien:
Resolves #1242


cppwfs:
LGTM

PR #1293, opened 2019-09-06
Application details page: fix unregister app redirection

### LINKED ISSUE #1291: Deleting an app does not redirect to apps page
When deleting an app, the user is not redirected to the apps page. Instead, (s)he stays on the app details page of the now non-existing app.

image

oodamien:
Resolves #1291


jvalkeal:
lgtm

PR #1302, opened 2019-09-24
Add import/export streams

### LINKED ISSUE #1292: Implement import/export UX for stream/task definitions
As a user, I'd like to have an option in the UI to select individual and/or all the available stream or task definitions, so I can export and import them to another SCDF instance.

**Notes:**
- Though this requires a backend API to fetch and massage the definitions, it would still be beneficial to explore the UX design as the first step
- It'd be useful to "define" what the exported format should be; is it a plain text file, or something UI-friendly, like JSON?
- While importing the exported definitions onto a different SCDF instance, we could come up with UX to incrementally validate the definitions

oodamien:
Resolves #1292

### DIFF
--- a/ui/src/app/shared/components/blocker/blocker.service.ts
+++ b/ui/src/app/shared/components/blocker/blocker.service.ts
@@ -1,4 +1,5 @@
 import { Injectable, Renderer2, RendererFactory2 } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
 
 /**
  * Confirm Service.
@@ -13,6 +14,8 @@ export class BlockerService {
    */
   private renderer: Renderer2;
 
+  public changeState: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
+
   /**
    * Constructor
    * @param rendererFactory
@@ -26,6 +29,7 @@ export class BlockerService {
    */
   lock() {
     this.renderer.addClass(document.body, 'blocked');
+    this.changeState.next(true);
   }
 
   /**
@@ -33,6 +37,7 @@ export class BlockerService {
    */
   unlock() {
     this.renderer.removeClass(document.body, 'blocked');
+    this.changeState.next(false);
   }
 
 }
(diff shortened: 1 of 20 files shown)


jvalkeal:
Create `ticktock = time | log', export without deploy:
```
{
  "date": 1569403539506,
  "streams": [
    {
      "force": false,
      "name": "ticktock",
      "dslText": "time | log",
      "originalDslText": "time | log",
      "description": "",
      "status": "undeployed"
    }
  ]
}
```

deploy and then export:
```
{
  "date": 1569403584945,
  "streams": [
    {
      "force": false,
      "name": "ticktock",
      "dslText": "time --spring.rabbitmq.addresses=rabbitmq-server:5672 | log --spring.rabbitmq.addresses=rabbitmq-server:5672",
      "originalDslText": "time | log",
      "description": "",
      "status": "deployed"
    }
  ]
}
```

When restoring we should use `originalDslText` to set the dsl as `dslText` text is reflected for runtime changes. We can think about later if we want to allow to choose which one to pick.

jvalkeal:
I'm also thinking if this exchange format should be somewhat official instead of just using all these fields as is. Logic might also live on a server side so that UI/CLI/REST would not need to worry about it to wrap to individual calls.

jvalkeal:
One other problem is these sensitive values, ie with:
```
stream1=time | jdbc --url=jdbc:mariadb://localhost:3306/dataflow
```
we get:
```
{
  "date": 1569485319872,
  "streams": [
    {
      "force": false,
      "name": "stream1",
      "dslText": "time | jdbc --url='******'",
      "originalDslText": "time | jdbc --url='******'",
      "description": "",
      "status": "undeployed"
    }
  ]
}
```
so import would not work.

jvalkeal:
Maybe we need to do a new import/export API's which can then bypass all logic around sensitive values.

ilayaperumalg:
The problem though is that we can't send the sensitive values as the REST response even if we enable the REST API to serve these values as is (without masking).

Typically, the sensitive values like these are bound to change between export and subsequent import and hence require changes. I think it would be ok if we let the users edit those sensitive values after re-import. This way we should be fine with the current implementation.

jvalkeal:
Merged per 7464ceaebe298a7bae84f26a2b904c78fb02a1cb

PR #1307, opened 2019-09-30
Revisit wording on Tasks/streams utils

### LINKED ISSUE #1305: Add note on handling sensitive properties during export and import
When handling the stream/task export, the associated sensitive properties are masked and the users need to be notified from the UI having a note like `The exported file needs to be modified for sensitive properties before importing`.

oodamien:
Resolves #1305

### DIFF
--- a/ui/src/app/tasks/tasks-utils/tasks-import/tasks-import.component.ts
+++ b/ui/src/app/tasks/tasks-utils/tasks-import/tasks-import.component.ts
@@ -42,7 +42,7 @@ export class TasksImportComponent {
    * Import options
    */
   options = {
-    optimize: false
+    excludeChildren: true
   };
 
   /**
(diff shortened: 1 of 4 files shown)


ui/src/app/tasks/tasks-utils/tasks-import/tasks-import.component.ts
```diff
@@ -42,7 +42,7 @@ export class TasksImportComponent {
    * Import options
    */
   options = {
-    optimize: false
+    excludeChildren: true
```
jvalkeal: Wrong way around. Now you have a checkbox enabled to exclude but childs are not excluded. Not sure why we even want this?

jvalkeal:
If you actually deselect a box, result is this:
tasks-import-error

jvalkeal:
file I used:
```
{
  "date": 1569858490325,
  "tasks": [
    {
      "name": "task1",
      "dslText": "t1: timestamp && t2: timestamp",
      "composed": true,
      "status": "UNKNOWN",
      "description": ""
    },
    {
      "name": "task1-t1",
      "dslText": "timestamp",
      "composed": false,
      "status": "UNKNOWN",
      "description": ""
    },
    {
      "name": "task1-t2",
      "dslText": "timestamp",
      "composed": false,
      "status": "UNKNOWN",
      "description": ""
    }
  ]
}
```

PR #1335, opened 2019-11-30
Track platform info in Audit history

sylank:
As a user, while launching Tasks against different platform backends (cf, k8s, etc.) in v2.0, I can't seem to find out where the actual execution occurred. It'd be good to keep track of the platform where the Tasks were launched, and surface that information in the UI.

Resolves: https://github.com/spring-cloud/spring-cloud-dataflow/issues/2891

image


jvalkeal:
lgtm, merged and squashed per 44ea28c50bcd207b08a9d23747c708265b314884

PR #1344, opened 2019-12-08
Flo new UI

oodamien:
**Revisit:**
Stream Flo creation
Stream Flo view
Task Flo creation
Task Flo view
Application properties


jvalkeal:
```
ticktock = time | log
```
Open options for `time` and these requests are made 5 times:
```
[HPM] GET /streams/definitions/ticktock -> http://localhost:9393/
[HPM] GET /streams/definitions/ticktock -> http://localhost:9393/
[HPM] GET /streams/definitions/ticktock -> http://localhost:9393/
[HPM] GET /streams/definitions/ticktock -> http://localhost:9393/
[HPM] GET /streams/definitions/ticktock -> http://localhost:9393/
```
Doesn't happen without name:
```
time | log
```

jvalkeal:
Not sure if this is for a future work but links have `properties` popup with empty input/output channels
flo-link-dummy-channels

jvalkeal:
Adding destination don't always update dsl(thought seems to be same issue in old as well):
flo-destination-update-bug

oodamien:
> requests are made 5 times

Flo Side (`Properties.Validators.uniqueResource`)

BoykoAlex:
> Not sure if this is for a future work but links have `properties` popup with empty input/output channels
Links have properties for the future multiple ports... Perhaps if there is only one input and one output port the  little properties icon won't be displayed... Will see how this multiple input/output feature shapes up in the near future...

BoykoAlex:
@jvalkeal I think it's ok to merge it for 2.4. The multiple input/output port work doesn't interfere with the current state of streams and composed tasks editors.
- The link properties icon will be gone for 2.4 - I'll roll it back to what it used to be
- The 4-5 requests for stream definition name when properties dialog pops up was around for some time unfortunately - something I'll look into

BoykoAlex:
@jvalkeal I have removed link properties icon and fixed the number of requests to check stream unique name. I think these were all the concerns for now... It seems like there are a few things to fix with the layout of flo editor and actions in SCDF UI... at least with resolution on my big screen... I think such problems we can fix gradually once this PR is in.

jvalkeal:
ok, lgtm for now, merging today.

jvalkeal:
squashed and merged per 979ae0500baa82ee93a4998a6585a2cb1dade411

PR #1352, opened 2019-12-17
Remove counters from Tasks

### LINKED ISSUE #1348: Remove the "count" from the tabs and sub-tabs
As a user, when I launch any of the list pages, I currently notice the lag due to the tabs/sub-tabs being rendered with the "count". The total counts of streams, tasks, or schedules are not usable metrics. They don't have to be queried every time I go to the list pages. 

The blocking calls to retrieve the results attribute towards the slower page load.

oodamien:
Resolves #1348


jvalkeal:
lgtm

jvalkeal:
Merged per 22775b7e8be1af33e90c1cfe0db039920c14aa46

PR #1357, opened 2020-01-08
Set proper link ends for Start and End nodes for text->graph

### LINKED ISSUE #1355: Outbound connections to the END node is unbalanced
On the latest UX changes, the connection ports don't appear to be balanced. 

image

BoykoAlex:
Fixes #1355


sabbyanandan:
@BoykoAlex: You may have addressed it globally, both for streams and tasks, but here's another on the composed task.

image

BoykoAlex:
@sabbyanandan @aclement alignment of `START` and `END` nodes is fixed as well now

PR #1358, opened 2020-01-09
Update Flo color shapes

### LINKED ISSUE #1356: Revisit the background color for the nodes
The background gray isn't very appealing. On a large monitor, it looks almost transparent; it might be tricky to use for some users, so let's revise the coloring to make it more neutral for the broader userbase.

Example:
image

oodamien:
Resolves #1356


BoykoAlex:
@oodamien I have pushed fixes for label truncation... Are we going to align labels around the shape's centre for the palette CT shapes too?

sabbyanandan:
I see this for streams:
image

And, tasks:
image

It seems alright to go with this for now. @BoykoAlex: If you are cool with it also, feel free to merge the PR.

BoykoAlex:
@sabbyanandan Are CT palette shapes ok? Or do we want the label to be centre aligned just like for CT shapes on the canvas?

sabbyanandan:
@BoykoAlex: Here's the current behavior:

- For streams, the palette and the nodes in the canvas, the text in both are left-aligned. 
- For tasks, the palette is left-aligned, but as you point out, the nodes in the canvas are center-aligned. 

It'd be nice to make them all look consistent (probably, center-aligned everywhere), but we can do this incrementally (maybe for 2.4 RC1?). Not a show stopper anyway.

BoykoAlex:
LGTM. Squashed and merged with 2e100d65d407e41f2fe933a54f1631374776e6bb

PR #1363, opened 2020-01-15
Add search on application properties modal

### LINKED ISSUE #1362: Add an option to search for deployer/app properties
As a user, while using the application or the deployer properties dialog window, I notice there are several 10s of properties, and it is hard to find the specific item by scrolling up and down. It'd be nice to search for a particular property with the help of a search bar on the top inside the dialog window.

**Acceptance:**
- Come up with a UX for searching for properties in the deployer or application properties window
- Searching for specific text will return context-specific results in the dialog window 
- Removing text from the search bar brings back all of the available properties for selection

oodamien:
Resolves #1362

### DIFF
--- a/ui/src/app/shared/flo/properties-groups/properties-groups-dialog.component.html
+++ b/ui/src/app/shared/flo/properties-groups/properties-groups-dialog.component.html
@@ -4,8 +4,14 @@
     <span aria-hidden="true">&times;</span>
   </button>
 </div>
-<form name="app-properties" (submit)="handleOk()" class="modal-body app-properties">
 
+<label class="modal-search">
+    <span class="search-wrapper">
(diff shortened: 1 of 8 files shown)


BoykoAlex:
@oodamien I have pushed fixes for label truncation... Are we going to align labels around the shape's centre for the palette CT shapes too?

sabbyanandan:
This PR only addresses the application properties; this is more so required for deployer properties, however. For instance, when a user opens the properties for K8s-deployer, there are 50+ entries that they can override.

Perhaps this PR is meant to only partially addressed for now. I just wanted to point this out, though.

BoykoAlex:
LGTM. Squashed and merged with b79ec9c0da68f977a9a8ca1ddbbd1703e3d413ff

PR #1383, opened 2020-02-19
Update size app_registration request (10.000)

### LINKED ISSUE #1381: Having more than 1000 entries in the table `app_registration` causes different errors
@xetra11 commented on [Fri Feb 07 2020](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3787)

**Description:**
We had more then 1000 apps registered. It seems that the SCDF frontend is only fetching 1000 and no more. We truncated the table and the issue was resolved.

**Release versions:**
springcloud/spring-cloud-dataflow-server:2.3.0.RC1

---

@sabbyanandan commented on [Wed Feb 12 2020](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3787#issuecomment-585474989)

Hi, @xetra11. That's incredible to hear that you have had >1000 apps in the registry. Curious whether if they were are all unique apps and what do the majority of the applications do from the business perspective? Anything that you can share would help reason through the requirement.

As for the limit, there's probably a hardcoded limit around, and that is not a good thing. We can look into loosening it up to make it dynamic, though.

I am looking forward to hearing your use-case requirements.

---

@jvalkeal commented on [Mon Feb 17 2020](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3787#issuecomment-587009831)

This is kind of a design flaw how apps are shown. Theoretically we should have designed every location in UI to use proper paging or not use paging on a rest api at all. Some part of an UI don't use paging and thus original workaround was to request one with size 1000.

It might be even better to fetch less and do multiple queries. We probably have this issue in other parts of an UI as well.

oodamien:
Resolves #1381

### DIFF
--- a/ui/src/app/apps/apps.workaround.service.ts
+++ b/ui/src/app/apps/apps.workaround.service.ts
@@ -115,7 +115,7 @@ export class AppsWorkaroundService {
 
   appVersions(appType: ApplicationType, appName: string): Observable<AppVersion[]> {
     return this.sharedAppsService
-      .getApps(new PageRequest(0, 1000), appType, appName, [{ sort: 'name', order: 'ASC' }, {
+      .getApps(new PageRequest(0, 10000), appType, appName, [{ sort: 'name', order: 'ASC' }, {
         sort: 'type',
         order: 'ASC'
       }])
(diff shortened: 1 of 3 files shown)


BoykoAlex:
LGTM. Rebased and merged with f8e4ad0bd84b097abdd0b9349efee786cb497b2b

Issue #1385, opened 2020-02-21
Poll for metrics only when the stream is expanded


sabbyanandan:
In the UI, when showing the list of streams, the call to get the status should not return the detailed metrics information for each stream. 

We can redesign this by,
- removing the 10s automatic polling by default
- poll for metrics on-demand and _only_ when the stream is expanded on the stream-list
- if all the streams are expanded, yes, we will poll for metrics for all the streams 

This requires a new method in scdf/skipper to return the detailed metrics. 

Depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/3800
Depends on https://github.com/spring-cloud/spring-cloud-skipper/issues/939

PR #1389, opened 2020-02-26
Optimize metrics calls on streams list

oodamien:
Part of #1385

### DIFF
--- a/ui/src/app/streams/streams/streams.component.ts
+++ b/ui/src/app/streams/streams/streams.component.ts
@@ -450,6 +450,7 @@ export class StreamsComponent implements OnInit, OnDestroy {
     }).filter((a) => a != null);
     this.itemsExpanded = value;
     this.updateContext();
+    this.loadStreamMetrics();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ui/src/app/streams/streams/streams.component.ts
```diff
@@ -458,7 +459,15 @@ export class StreamsComponent implements OnInit, OnDestroy {
   loadStreamMetrics() {
     const streamNames = this.streamDefinitions && Array.isArray(this.streamDefinitions.items) ?
       this.streamDefinitions.items
-        .filter(i => (i.status === 'deployed') || (i.status === 'deploying') || (i.status === 'partial'))
+        .filter(i => {
```
ilayaperumalg: This will still load stream metrics per page. What we need is to have this invoked only when the user clicks on the `show graph` link.

oodamien: ```if (streamNames.length) {```
This statement insure that we don't call the WS if there is no streams in the filtered array.

ilayaperumalg: ok, I understand it better now. Thanks @oodamien!

ilayaperumalg:
LGTM, merging.

PR #1400, opened 2020-03-25
Load parameters on re-launch a task

### LINKED ISSUE #1387: Show deployment properties from previous task execution in the Parameters field on task launch page
We have a scenario as described in issue https://github.com/spring-cloud/spring-cloud-dataflow/issues/3773 where users need to know what properties they will be using when they launch a task.  Hence we need to show them the deployment properties from the previous execution in the `Parameters` field on the task launch page.   The `deploymentProperties` can be obtained by calling the `tasks/executions/<task-execution-id>` using the `task-execution-id` field from the definitions request made when going to the task launch page.   
Make sure that if their is redacted data i.e. a value of `******` that we do not resend that data.

oodamien:
Resolves #1387


cppwfs:
LGTM

PR #1478, opened 2020-07-06
Move apps metadata from Element#attr(...) to Element#prop(...)

BoykoAlex:
Moving stream app shapes metadata from `Element.attr(...)` into `Element.prop(...)` (or `Element.set(...)`)
`Element` object is being cloned by JointJS core during various operations properties set with `Element.attr(...)` are being cloned while properties set with 'Element.prop(...)` or Element.set(...) not deep cloned but rather assigned. This allows to have methods and keep cached promises.
JoinJS folks recommended to use `Element.attr(...)` only for SVG dom element attributes and keep the "business" data under `Element.prop(...)'

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -46,7 +46,7 @@
     "npm-check-updates": "^4.1.0",
     "rxjs": "6.5.4",
     "sockjs-client": "1.4.0",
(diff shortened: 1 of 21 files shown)


ilayaperumalg:
LGTM, merging.

PR #1486, opened 2020-07-09
Add platforms support for schedules

### LINKED ISSUE #1433: UI: Support platform selection for schedules
When more than platform is available a user must select a platform when 
- [ ] Creating a schedule
- [ ] Deleting a schedule
- [ ] Listing schedules
- [ ] Obtain a list of platforms that have schedules enabled via the 
```
    "tasks/platforms/scheduler": {
      "href": "http://localhost:9393/tasks/platforms/scheduler"
    },
```
Depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/3975

More can be read here: https://docs.google.com/document/d/17G302I3Z-w_74umJDmTf9IiXV3jgu5LKwMyIt-wjAK0/edit#

oodamien:
Resolves #1433

### DIFF
--- a/ui/src/app/shared/model/platform.ts
+++ b/ui/src/app/shared/model/platform.ts
@@ -6,15 +6,15 @@
  */
 export class Platform {
 
-  public name: String;
+  public name: string;
 
-  public type: String;
+  public type: string;
 
-  public description: String;
+  public description: string;
 
   public options: Array<any>;
 
-  constructor(name: String = '', type: String = '', description: String = '', options: Array<any> = []) {
+  constructor(name: string = '', type: string = '', description: string = '', options: Array<any> = []) {
     this.name = name;
     this.type = type;
     this.description = description;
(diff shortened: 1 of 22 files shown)


cppwfs:
When creating a schedule an error occurs:
1. Go to task definition page
2. On the drop down select `Schedule Task`

The UI is attempting this call: `http://localhost:4200/tasks/schedules` , however the server returns this error message 
```
[{"logref":"IllegalArgumentException","message":"Must select a platform.  A default could not be determined because more than one platform had an associated scheduler"}]
``` 
In cases where there is more than one platform a user must select the platform when requesting schedules.

cppwfs:
Also seeing same error when going to the schedules list page.

jvalkeal:
It's weird that when I used server locally and just added `foo` and `bar` platforms for cf, I don't see these errors.

cppwfs:
Hmm...   is scheduler enabled for both platforms?

jvalkeal:
Does it check presence of a scheduler before trying to use those? It would make sense to target from local to cf but I was expecting some error to get thrown.

cppwfs:
Dataflow checks to see if a task platform TaskLauncher has a scheduler associated with it.  If the tasklauncher has a scheduler associated with it, dataflow adds that to the pool of available schedulers.    If more than one tasklauncher has a scheduler then, dataflow will throw that exception.

cppwfs:
I kicked the tires on the PR against the latest SCDF and it looks good.

PR #1506, opened 2020-07-23
Tap should have one output port

BoykoAlex:
Fix tap shape port. Currently it has an input port instead of an output port
DSL: `:s1.something > jdbc`


ilayaperumalg:
LGTM, merging.

PR #1532, opened 2020-08-14
Schedule Platform Integration

oodamien:


cppwfs:
I'm getting this exception on the server when trying to schedule:  `The platform undefined does not exist`
I think the word `undefined` is being passed instead of the platform.

cppwfs:
LGTM

cppwfs:
Squashed Merged.

PR #1569, opened 2020-10-05
Fix job execution link to step

### LINKED ISSUE #1568: The link for the step execution id uses the job instance id instead of the job execution id
On the job executions detail page `http://localhost:9393/dashboard/#/tasks-jobs/job-executions/<job-execution-id>`  a user can click the step execution id to go the step-execution detail page for the the step-execution-id.    However the job-execution-id portion of the URL is the job-instance-id, when it should be the job-execution-id.   

i.e. ` http://localhost:9393/dashboard/#/tasks-jobs/job-executions/<job-execution-id>/<step-execution-id>`

oodamien:
Resolves #1568


ilayaperumalg:
LGTM, merging.

PR #1585, opened 2020-10-27
Revisit default size column datagrid

### LINKED ISSUE #1582: Fix truncate issue on dsl text/datagrid and datagrid col size

oodamien:
Resolves #1582


jvalkeal:
lgtm

PR #1609, opened 2020-12-01
Hide instance count option in Stream Update

### LINKED ISSUE #1525: Hide instance count option in Stream Update
Changing application instance count isn't supported in `Stream Update` and hence, the option to specify the `instance count` in Stream Update should be hidden.

oodamien:
Resolves #1525


jvalkeal:
lgtm

PR #1617, opened 2020-12-15
Adopt ESLint

### LINKED ISSUE #1612: Migrate Linter (deprecated)

oodamien:
Resolves #1612


jvalkeal:
I was thinking that we could try to do this change right before next GA release as that makes it easier to backport things.

sabbyanandan:
We will have to resubmit a new PR.

Issue #1620, opened 2020-12-19
Jobs executions spins infinitely on NoSuchTaskBatchException


snimmagadda1:
Did some searching and didn't see reference to this so apologies if this is duplicate or expected behavior.. 

I've run into this on a couple different dataflow servers in which an app or task gets imported and run, then the job executions tab becomes unusable with the following 404 for `jobs/thinexecutions?page=0&size=20`

```
{
  "_embedded": {
    "errors": [
      {
        "message": "No corresponding taskExecutionId for jobExecutionId 1.  This indicates that Spring Batch application has been executed that is not a Spring Cloud Task.",
        "logref": "NoSuchTaskBatchException",
        "_links": {
          "self": {
            "href": "/"
          }
        }
      }
    ]
  }
}
```

The only resolution I've found is to manually go into the DB and delete or clear the job history in order for that screen to become usable again (perhaps this is also doable via the shell also). Is there a provided way to correct this on the UI? Maybe the messaging around it could be clearer on the UI.

<img width="1127" alt="Screen Shot 2020-12-18 at 11 02 53 PM" src="https://user-images.githubusercontent.com/19558427/102681325-97085800-418e-11eb-9fc1-d176e8675c21.png">

jvalkeal:
Maybe we can do a bit more on a server side but UI side bug I'd say is the spinner and missing error handling.

cppwfs:
This needs to be backported to 3.0.x line.

cppwfs:
Backported.

PR #1622, opened 2020-12-23
Add messaging when error from thinexecutions endpoint on Job executions screen

snimmagadda1:
Addresses #1620.

Came across this. When a job isn't properly tagged as a Task app, it causes a 404 which causes the job executions screen to spin without any prompts to the user. This can be a bit misleading, so this change adds a notification on the screen when this happens. 

Screen Shot 2021-01-05 at 11 40 30 AM

### DIFF
--- a/ui/src/app/tasks-jobs/jobs/jobs.component.ts
+++ b/ui/src/app/tasks-jobs/jobs/jobs.component.ts
@@ -38,6 +38,8 @@ export class JobsComponent extends DatagridComponent {
           this.page = page;
           this.updateGroupContext(params);
           this.loading = false;
+        }, error => {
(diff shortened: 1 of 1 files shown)


oodamien:
Hi @snimmagadda1,
Thanks for the fix. 
Can you turn false the loading state also?

snimmagadda1:
> Can you turn false the loading state also?

@oodamien I can, but I think since the API call fails w/ 404 the screen won't list the proper executions because no results are being returned from the backend...

oodamien:
@snimmagadda1 Yes, at least the user can click on the refresh button. We don't handle an error state in this list pages so far.

PR #1626, opened 2021-01-06
Overhaul task launch

### LINKED ISSUE #1515: Redo task-launch screen to follow similar UX as streams

jvalkeal:
- Implement new task launch page in a similar way
  than stream deploy.
- Add sections for crt app options.
- Add section arguments.
- Normal task vs. composed task is sligthly different
  as ie we don't show ctr options.
- Needs a spesific changes on a server side.
- Fixes #1515
- There's a few follow-up issues to hand after this

You need to run server PR spring-cloud/spring-cloud-dataflow#4314 manually until it's merged

### DIFF
--- a/ui/src/app/flo/task-flo.module.ts
+++ b/ui/src/app/flo/task-flo.module.ts
@@ -46,7 +46,8 @@ import { SharedModule } from '../shared/shared.module';
   ],
   exports: [
     ViewComponent,
-    TaskFloCreateComponent
+    TaskFloCreateComponent,
+    TaskPropertiesDialogComponent
   ]
 })
 
--- a/ui/src/app/shared/api/task.service.ts
+++ b/ui/src/app/shared/api/task.service.ts
@@ -1,13 +1,14 @@
 import { Injectable } from '@angular/core';
 import { HttpClient, HttpParams } from '@angular/common/http';
+import { forkJoin, Observable } from 'rxjs';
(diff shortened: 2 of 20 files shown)


cppwfs:
For composed task properties we need to include the composed task definition name after the `deployer.` or `app.`.   For example the UI is rendering `app.t1.timestamp.format=YYYY`  .  It should be `app<myCtrDef>.t1.timestamp.format=YYYY`

jvalkeal:
Nope, that's what I removed so that actual launch properties don't need to know under what definition name it belongs to. This way if you have same exact task definition just with different name, you can copy/paste everything.

cppwfs:
Hmm. this may be a CTR side issue.   I just ran the following: 
image
The following post was executed against dataflow: `http://localhost:4200/tasks/executions?name=ctr0106&properties=app.t1.timestamp.format=YYYY`
But when looking at the resulting launch of CTR  task launch I see the property `app.t1.timestamp.format=YYYY`.   But the `t1` task execution for the task that was launched does not show the property.

cppwfs:
But passing `app.ctr0106.t1.timestamp.format=YYYY` works.

jvalkeal:
Ah I'll look that, got confused between app properties vs. app arguments.

oodamien:
LGTM

cppwfs:
LGTM.

cppwfs:
Rebased, squashed, merged.

PR #1636, opened 2021-01-13
FIx builder deploy validation

### LINKED ISSUE #1631: Mandatory field is not marked as required when deploying streams
I have 2 platform accounts configured in SCDF; however, when deploying the stream, I get a blanket error popup and it doesn't point to what exactly is missing.

image

When I select the platform dropdown, and I hit the deploy button again, I get to see a red highlight, but it is not fully visible.

image

oodamien:
Resolves #1631


jvalkeal:
lgtm

PR #1638, opened 2021-01-16
Fix for search for disabled features

### LINKED ISSUE #1629: Search fails to load tasks when features are disabled
When some dataflow features are disabled and I try to search for a task, the search will not show the task.

**Reproduce:** 

```
 version: '3'

services:
  mysql:
    image: mysql:5.7.25
    container_name: dataflow-mysql
    ports:
      - "3306:3306"
    environment:
      MYSQL_DATABASE: dataflow
      MYSQL_USER: root
      MYSQL_ROOT_PASSWORD: rootpw
    expose:
      - 3306
      
  dataflow-server:
    image: springcloud/spring-cloud-dataflow-server:2.7.0
    container_name: dataflow-server
    ports:
      - "9393:9393"
    environment:
      - SPRING_DATASOURCE_URL=jdbc:mysql://mysql:3306/dataflow
      - SPRING_DATASOURCE_USERNAME=root
      - SPRING_DATASOURCE_PASSWORD=rootpw
      - SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.mariadb.jdbc.Driver
      - SPRING_CLOUD_DATAFLOW_FEATURES_STREAMS_ENABLED=false
      - SPRING_CLOUD_DATAFLOW_FEATURES_SCHEDULES_ENABLED=false
      - SPRING_CLOUD_DATAFLOW_FEATURES_TASKS_ENABLED=true
      - SPRING_DATASOURCE_INITIALIZATION_MODE=always
    entrypoint: "./wait-for-it.sh mysql:3306 -- java -jar /maven/spring-cloud-dataflow-server.jar"
    volumes:
      - ${HOST_MOUNT_PATH:-.}:${DOCKER_MOUNT_PATH:-/root/scdf}
      - /var/run/docker.sock:/var/run/docker.sock
      
  app-import:
    image: springcloud/openjdk:2.0.0.RELEASE
    container_name: dataflow-app-import
    depends_on:
      - dataflow-server
    command: >
      /bin/sh -c "
        ./wait-for-it.sh -t 180 dataflow-server:9393;
        wget -qO- 'http://dataflow-server:9393/apps' --post-data='uri=${TASK_APPS_URI:-https://dataflow.spring.io/task-maven-latest&force=true}';
        echo 'Task apps imported'"
```  
With the following docker-compose one can start a simple dataflow stack.
     

1. Create a new task in the ui or with the shell (I created one named abc)
2. Search for the task in the searchbar ( The searchbar will show the tasks header but not show any task )

    
**Issue:** 

In my point of view the search code always except that streams and tasks are enabled. It does not validate in the frontend or backend if any feature is missing or the objects are null.

**Solution:** 

1. A quick and dirty solution would be, that the frontend uses the Safe Navigation Operator to validate that the stream object is null. This would result in the tasks been shown but still the frontend would throw errors because the streams could not be get.

2. A quite better option would be to know which features are enabled. Somehow the about page retrieves if streams or tasks are enabled and maybe this information could be used to not query tasks or streams at all if they are disabled.

hpoettker:
fixes #1629

### DIFF
--- a/README_DEV.md
+++ b/README_DEV.md
@@ -54,7 +54,7 @@ The following commands are available:
 $ npm start
 
 # Run the unit tests
-$ ng test --single-run
+$ ng test --watch=false
 
 # Run lint
 $ ng lint
--- a/ui/src/app/shared/component/search/search.component.html
+++ b/ui/src/app/shared/component/search/search.component.html
@@ -49,8 +49,8 @@
         <h4>Task(s)</h4>
         <ul>
           <li *ngFor="let task of results.tasks.items;let i = index"
-              [class.active]="selected === (i + results.apps.items.length + results.streams.items.length)">
-            <a (mouseover)="selected = (i + results.apps.items.length + results.streams.items.length)"
+              [class.active]="selected === (i + results.apps.items.length + (results.streams ? results.streams.items.length : 0))">
+            <a (mouseover)="selected = (i + results.apps.items.length + (results.streams ? results.streams.items.length : 0))"
                (click)="navigate('/tasks-jobs/tasks/' + task.name)">
               <span class="text-truncate">
                 {{task.name}} <span class="{{task.labelStatusClass()}}">{{task.status}}</span>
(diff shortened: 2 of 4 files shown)


sabbyanandan:
Thanks for attempting to fix this workflow, @hpoettker. We will have a look and get back to you with updates soon.

PR #1656, opened 2021-02-12
Delay loading ctr options

### LINKED ISSUE #1644: Make composed task launch page to load faster
When launching ctr we need options for ctr app itself and currently this is not done asynchronously, thus you may see lag of page to load especially it it needs to get loaded from repository. Should do same tricks as stream deploy where things are implemented in more user friendly manner.

jvalkeal:
- As loading ctr options may be a tad slow as it backing
  metadata may come from a jar needed to get downloaded and
  may result error, etc. Delay this operation after
  initial form is built and then update needed form states.
- Also skip load if task is not ctr.
- Fixes #1644

NOTE: for existing functionality to see behaviour, just restart a server and zap ~/.m2/repository/org/springframework/cloud/spring-cloud-dataflow-composed-task-runner.

### DIFF
--- a/ui/src/app/shared/model/task.model.ts
+++ b/ui/src/app/shared/model/task.model.ts
@@ -98,7 +98,13 @@ export class TaskLaunchConfig {
   }[];
 
   /** Ctr options metadata */
-  ctr: ValuedConfigurationMetadataProperty[];
+  ctr: {
+    options: ValuedConfigurationMetadataProperty[];
+    optionsState: {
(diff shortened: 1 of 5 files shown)


cppwfs:
LGTM
Kicked the tires with 
* existing CTR2.8 jar
* cleared CTR2.8 jar forcing a load
* Network turned off CTR2.8 jar with no existing CTR2.8 jar
* Task No CTR launch

PR #1659, opened 2021-02-16
Redirect to task execution after launch

### LINKED ISSUE #1599: As a user I would like to be redirect to the task execution when triggering a task
Hi,

after triggering a task via the ui at tasks/definitions/my-task/summary, right now there is a redirect to the task overview at dashboard/#/tasks/definitions. 

I assume that for most usecases it would be great to be redirected to the execution of the started task dashboard/#/tasks/executions/1234

Could you improve the ui in such a way? Will save three clicks each time ;) Might even make sense to do so after triggering a task from the task overview at dashboard/#/tasks/definitions.

jvalkeal:
- Now simple parse body response from a task launch
  and redirect to returned task execution.
- No special handling of batch
  (aka trying to redirect job execution instead) as batch
  is expected to be a task?
- Fixes #1599

### DIFF
--- a/ui/src/app/shared/api/task.service.ts
+++ b/ui/src/app/shared/api/task.service.ts
@@ -71,7 +71,7 @@ export class TaskService {
     return forkJoin(tasks.map(task => this.destroyTask(task)));
   }
 
-  launch(taskName: string, args: string, props: string): Observable<any> {
+  launch(taskName: string, args: string, props: string): Observable<number> {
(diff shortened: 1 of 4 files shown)


cppwfs:
To answer the question on the second bullet point, Batch in dataflow must always be a task.

cppwfs:
LGTM
Navigated both CTR and Task apps.

PR #1665, opened 2021-02-19
Fix showing tasks platform properties

### LINKED ISSUE #1661: Task execution: show Platform Properties is broken
Running a task and then navigate to the execution.
The property Platform Properties has an odd render.

image

jvalkeal:
- Fix how platform properties are shown with executions for
  task/batch as this broken in templates when underlaying
  structure were changed to be more type safe.
- Fixes #1661


cppwfs:
lgtm

PR #1672, opened 2021-02-24
Adds platform information to schedule lists and task definition detail

### LINKED ISSUE #1670: Platform name not displayed with Task Schedules
Schedule entries don't display the platform name(they report only N/A) on schedules page.
The also don't show up with the schedules on the Task Definition Detail page.

Refer to issue: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4369

cppwfs:
resolves #1670

### DIFF
--- a/ui/src/app/shared/api/schedule.service.ts
+++ b/ui/src/app/shared/api/schedule.service.ts
@@ -37,6 +37,12 @@ export class ScheduleService {
       .get<any>(`${url}?platform=${platform}`)
       .pipe(
         map(SchedulePage.parse),
+        map((page: SchedulePage) => {
+          page.items.forEach(schedule => {
+            schedule.platform = platform;
+          });
+          return page;
+        }),
         catchError(ErrorUtils.catchError)
       );
   }


ui/src/app/shared/api/schedule.service.ts
```diff
@@ -37,6 +37,12 @@ export class ScheduleService {
       .get<any>(`${url}?platform=${platform}`)
       .pipe(
         map(SchedulePage.parse),
+        map((page: SchedulePage) => {
```
jvalkeal, line 40: That parse method already tries to use platform so it might work better like this:
```
      .pipe(
        map((input) => {
          input.platform = platform;
          return input;
        }),
        map(SchedulePage.parse),
        catchError(ErrorUtils.catchError)
      );

```

jvalkeal: Though I kinda hate to modify input like this but we always get to these issues when using `any` type which makes contract between server response vs. data in a page a bit obscure.

cppwfs:
I understand, the concern.  
I did try the code update above but the second map overwrites the platform :-(

jvalkeal:
Ah bummer, I was looking wrong parse method in `schedule.model.ts`. We're not properly mocking `schedule.server.ts` so there's no tests for these methods.

jvalkeal:
Let's go

PR #1681, opened 2021-03-03
Add modal to task launch

### LINKED ISSUE #1680: Launching Tasks from Builder does not grey out screen
In the past when launching tasks (especially on Cloud Foundry) the screen would go grey and show the cycle icon until the server acknowledged the launch, thus preventing the user from clicking the "Launch Task" button again.   Launching from the builder now does not grey out the screen and looks like the "Launch Task" button was not clicked, so user clicks again getting a second launch.

jvalkeal:
- Same way as stream deploy, blocking modal
  when launch happens.
- Fixes #1680


cppwfs:
LGTM

PR #1687, opened 2021-03-14
Show horizontal scroll for logs not only at bottom

hpoettker:
The logs available in SCDF can contain quite long lines. Currently, the horizontal scroll bar for that is only visible when one scrolls to the very bottom of the log. This makes it unnecessarily complicated to read the long lines, in my opinion.

Clarity's `modal-body` sets `max-height` to `70vh`. Only if this is inherited for the log, the horizontal scroll bar is always shown.

I observed the problem in both Firefox and Chrome. The PR fixes it for both. The only other use of `clr-log` is for the exit description of steps. With the PR applied, that widget looks same as before in both browsers, respectively.

Below two screenshots from Firefox with the PR applied:

task_log
step_exit_desc

### DIFF
--- a/ui/src/scss/styles.scss
+++ b/ui/src/scss/styles.scss
@@ -65,9 +65,11 @@ h1 {
 
 .clr-log {
   padding: 1rem 0;
+  max-height: inherit;
 
   pre {
     overflow: scroll;
+    max-height: inherit;
     background: var(--clr-header-6-bg-color, #00364d);
     padding: 1rem;
     color: var(--clr-header-title-color, #fafafa);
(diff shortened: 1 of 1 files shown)


hpoettker:
For short content like `N/A`, the widget for the step exit description currently looks different in Chrome and Firefox. In Firefox, it looks as in the above screenshot, i.e. without any scrollbars. In Chrome, it looks like the screenshot below.

I added a commit to set the scroll behavior to `auto`. Then it looks as above in both browsers, i.e. without scroll bars for `N/A`. The log of the task execution looks unchanged.

step_exit_desc_chrome

jvalkeal:
Thanks, looks good!

PR #1689, opened 2021-03-21
Split task arguments only at first equal sign

hpoettker:
When a user wants the sub tasks of a composed task to be started with the argument `foo=bar`, they can start the composed task with the argument `--composed-task-arguments=foo=bar`. On the page for the execution of the composed task, the `=bar` is currently scrapped as seen here:

without_bar

With the proposed change, it looks like this

with_bar


jvalkeal:
I think this is a good fix/hack as we anyway use that as array of size 2 in templates.

Issue #1686, opened 2021-03-12
User can not enter a command line argument without a following = symbol when launching tasks.


cppwfs:
Validation preventing a user from entering:
myarg
--myarg2

These are valid commandline args.

cppwfs:
Refer to issue https://github.com/spring-cloud/spring-cloud-dataflow/issues/4427  for more detail.

PR #1691, opened 2021-03-23
Fix line breaks in textarea for Firefox

hpoettker:
In Firefox, line breaks are not shown in the `textarea` widgets when launching tasks. It currently looks as in the screenshot below. In Chrome, everything looks fine.

The PR changes `white-space` from `nowrap` to `pre`. That way, in both browsers
- line breaks are shown,
- and long lines are not wrapped

I also set `overflow-x` from `scroll` to `auto` to only show the horizontal scroll bar when there is actually something to scroll.

That the arguments are flagged as invalid in the screenshot is a separate issue: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1686

without_fix

### DIFF
--- a/ui/src/app/tasks-jobs/tasks/launch/free-text/free-text.component.scss
+++ b/ui/src/app/tasks-jobs/tasks/launch/free-text/free-text.component.scss
@@ -38,9 +38,9 @@
 
   textarea {
     margin: 0;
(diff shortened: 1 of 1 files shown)


oodamien:
LGTM
Thanks @hpoettker

PR #1697, opened 2021-04-02
Add platform to navigate to the schedule details

### LINKED ISSUE #1671: Task Schedule Detail can only viewed for one of the platforms
**[To Reproduce]**
Create 2 schedules 1 for default (sched-a) and 1 for practice (sched-b).
On either the task schedule page or the task definition details page select sched-b and the detail page will appear, but get an error when trying to get to sched-a.   The error reported in the browser console is: 
image

**[Cause]**
The UI does not pass the platform that is associated with schedule when requesting the detail i.e. it requests `http://192.168.99.103:30152/tasks/schedules/task-sched-a`.   It should request `http://192.168.99.103:30152/tasks/schedules/myschedc?platform=default`

To answer the question, why does it work in one case and not the other.   This is because the server will use the last platform in the list to search for a schedule detail if the platform is not specified.    

Task Detail also does not show platform.
Refer to issue: spring-cloud/spring-cloud-dataflow#4369

oodamien:
Resolves #1671

### DIFF
--- a/ui/src/app/shared/api/schedule.service.ts
+++ b/ui/src/app/shared/api/schedule.service.ts
@@ -47,15 +47,16 @@ export class ScheduleService {
       );
   }
 
-  getSchedule(scheduleName: string): Observable<Schedule> {
+  getSchedule(scheduleName: string, platformName?: string): Observable<Schedule> {
     // return of(GET_SCHEDULE)
     //   .pipe(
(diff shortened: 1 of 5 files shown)


cppwfs:
LGTM

cppwfs:
Rebased, Merged.

PR #1717, opened 2021-05-25
Update helper description on Schedule creation page

### LINKED ISSUE #1716: Schedule page properties should show prefix as scheduler instead of deployer
Screenshot 2021-05-25 at 3 56 59 PM

Along with `app` prefix we need to show `scheduler` as a prefix instead of `deployer`.

oodamien:
Resolves #1716


ilayaperumalg:
LGTM, merging.

PR #1719, opened 2021-05-25
Update conditions on properties modal filtering

### LINKED ISSUE #1714: Properties modal filtering awkward to use
We should just make this filtering to work better so that we get all matches accurately:

deployer-props-filter

oodamien:
Resolves #1714


jvalkeal:
It works better with deployment platform properties, but not so well with app or ctr properties modal.

jvalkeal:
lgtm now

Issue #1721, opened 2021-05-26
Can't remove a task ctr property


jvalkeal:
When defining any props which goes between builder and freetext and we've dehydrated one, you can't remove it from a launch manifest by removing from field values. It works with freetext if value is effectively left empty.

PR #1722, opened 2021-05-26
Add migration concept for old ctr app properties

### LINKED ISSUE #1715: Handle old ctr props
With new task launch ui we moved to not require task name in props anymore so that those are portable between tasks.
```
app.t1.timestamp.format=yyyy
```
On a server we still handle old style:
```
app.ctr1.t1.timestamp.format=yyyy
```
Need to think how we represent this in UI.

jvalkeal:
- Add Some UI fu to let user to automatically migrate
  from old ctr app properties(given in 2.7.x line) and
  migrate those to new expected ctr options.
- For this work correctly also #1721 needs to be fixed which
  is a bigger topic and will get done in a separate commit.
- Fixes #1715

### DIFF
--- a/ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
+++ b/ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
@@ -95,6 +95,11 @@ export interface Builder {
   };
 }
 
+export interface Migrations {
+  migratedNomatch: string[];
+  migratedMatch: RegExpExecArray[];
+}
+
 @Component({
   selector: 'app-task-launch-builder',
   templateUrl: 'builder.component.html',
@@ -142,6 +147,8 @@ export class BuilderComponent implements OnInit, OnDestroy {
     arguments: true
   };
 
+  private migrations: Migrations;
+
   constructor(
     private taskLaunchService: TaskLaunchService,
     private changeDetector: ChangeDetectorRef,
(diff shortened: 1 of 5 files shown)


cppwfs:
Create task ctrx with the following definition `t1:timestamp&&t2:timestamp`
1) Select launch for `ctrx` 
2) Set the timestamp format property  for each of the tasks (t1 & t2).
3) Launch the tasks
4) Immediately go back to launch the task `ctrx`  and you will see the properties are not set.
5) Refresh the page and the properties are there.

ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
```diff
@@ -1022,4 +1030,53 @@ export class BuilderComponent implements OnInit, OnDestroy {
       this.launch.emit({props: this.getProperties(), args: this.getArguments()});
     }
   }
+
+  hasMigrations(): boolean {
+    return this.migrations && this.migrations.migratedMatch && this.migrations.migratedMatch.length > 0;
+  }
+
+  migrage() {
```
cppwfs, line 1038: migrate?

jvalkeal: hah, funny how you don't see these typos.

cppwfs: I hear ya'.   Done it more times that I can remember :-)

cppwfs:
LGTM

cppwfs:
Squashed Merged.

Issue #1730, opened 2021-05-27
Can't remove a task app property


jvalkeal:
Related as follow up from #1721. There's same issues with i.e. clearing out `timestamp.format` but it's more involved as I think ctr app needs to be modified as well.

cppwfs:
Moving this issue to https://github.com/spring-cloud/spring-cloud-dataflow/issues/5085.

PR #1731, opened 2021-05-27
Handle clearing out ctr options

### LINKED ISSUE #1721: Can't remove a task ctr property

jvalkeal:
- Now allows to clear out existing crt option which will
  then get removed from manifest by sending option as
  empty value.
- NOTE: this will only touch crt and we have same issues
  i.e. with app option, which goes to different issue #1730
- Fixes #1721


cppwfs:
LGTM.  Found some associated issues (not related to this PR) and opened up issues for those.

cppwfs:
Merged

PR #1739, opened 2021-06-02
Fix redirection on cancel task creation

### LINKED ISSUE #1736: Create Task dialog does not take you back to create page if invalid task def
If a user enters an invalid taskl definition they are allowed to select `Create Task` button and the Create Task dialog appears.  The user enters task name and description.   They then click the `Create The Task` button.  An error is displayed as expected.   At this point the user can click cancel but it takes them back to the task definition page.

Shouldn't the dialog take the user back to the task creation page?

oodamien:
Resolves #1736


cppwfs:
LGTM

PR #1750, opened 2021-06-28
Force only default version on applications quick search

### LINKED ISSUE #1749: Quick search returns duplicate applications
With 2 application versions, the quick search should only return the default version.

oodamien:
Resolves #1749


jvalkeal:
lgtm

PR #1754, opened 2021-08-11
Show id instead of name on App view

oodamien:
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/3815


jvalkeal:
for app view, lgtm. app options in deploy form is then a separate ticked.

jvalkeal:
Put wrong issue to commit as it was on dataflow side. Moved issue here per #1755

PR #1758, opened 2021-08-18
Group properties on App Properties Modal

oodamien:


sabbyanandan:
@BoykoAlex: Please review and get this merged.

dturanski:
Grouping looks great.  Not sure about the top level group `>` . do we need that? The search does not work AFAICT.

dturanski:
Some properties are missing.  e.g., in Tasklauncher Sink. Missing TaskLauncherFunctionProperties and TriggerProperties. I will verify it's in the metadata. https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-tasklauncher/spring-cloud-dataflow-tasklauncher-sink/src/main/resources/META-INF/dataflow-configuration-metadata.properties

BoykoAlex:
I think top level group could have a title **Main** or **General** or **Common** or perhaps just be outside any group. Those are the properties added from the UI (i.e. label, stream name for some elements). There is usually just a few such properties thus they could live outside groups.
Filter seemed to work for me though...
Checked props for 2 stream apps - they seem to be all there.

If there is only one "main" group present think it'd be best not to have a group shown in the UI - just minor enhancement. If too painful to implement can be done in the future if more people find it as a good thing to do

BoykoAlex:
LGTM. Merged in 15436483680ad1de7896d7942e4da6d8e8dbd260 (forgot to squash commits)

While testing I felt that may need to do something with groups when property filter text is entered. Don't think we'd want to expand groups to let user see what matched the filter as this might just defeat the purpose of grouping. However, show the number of properties in the group may help when filtering is active. It would help the user not to expand empty groups to look for properties :-) Other option could hide empty groups... but think showing number of properties in a group may just do the trick.

PR #1760, opened 2021-08-27
Single Step Job Feature [DRAFT]

oodamien:


sabbyanandan:
Woot, woot, this is making progress!! 😍

BoykoAlex:
I have integrated this work with the backend PR: https://github.com/spring-cloud/spring-cloud-dataflow/pull/4671

Therefore, once the PR applied to SCDF server code:
- `./mvnw clean install` in server code root folder
- Run the server
- Register the `ssbj` task from maven coordinates URI: https://github.com/spring-cloud/spring-cloud-dataflow/pull/4671/files

Now, the UI with proper content for readers and writers tabs should be available for the task **ssbj**. The tabs would show for **timestamp** as well currently but we should look into in displaying tabs only if app's metadata has any `optionGroups` set.

Issue #747, opened 2018-04-24
Setting server context-path breaks dashboard


sabbyanandan:
_From @EHJ-52n on April 24, 2018 8:52_

### Problem

I am using the docker image `springcloud/spring-cloud-dataflow-server-local:1.4.0.RELEASE` and set the following environment configuration in my docker-compose.yml for this image:
```yaml
environment:
  - spring.cloud.dataflow.applicationProperties.stream.spring.rabbitmq.host=rabbitmq
  - server.contextPath=/dataflow
  - spring.cloud.config.enabled=false
```
The result is, that I can access the server via `http://localhost:9393/dataflow/`.

The response looks like:
```json
{
  _links: {
    dashboard: {
      href: "http://localhost:9393/dataflow/dashboard"
    },
    streams / definitions: {
      href: "http://localhost:9393/dataflow/streams/definitions"
    },
    streams / definitions / definition: {
      href: "http://localhost:9393/dataflow/streams/definitions/{name}",
      templated: true
    },
    streams / deployments: {
      href: "http://localhost:9393/dataflow/streams/deployments"
    },
    streams / deployments / deployment: {
      href: "http://localhost:9393/dataflow/streams/deployments/{name}",
      templated: true
    },
    runtime / apps: {
      href: "http://localhost:9393/dataflow/runtime/apps"
    },
    runtime / apps / app: {
      href: "http://localhost:9393/dataflow/runtime/apps/{appId}",
      templated: true
    },
    runtime / apps / instances: {
      href: "http://localhost:9393/dataflow/runtime/apps/{appId}/instances",
      templated: true
    },
    metrics / streams: {
      href: "http://localhost:9393/dataflow/metrics/streams"
    },
    tasks / definitions: {
      href: "http://localhost:9393/dataflow/tasks/definitions"
    },
    tasks / definitions / definition: {
      href: "http://localhost:9393/dataflow/tasks/definitions/{name}",
      templated: true
    },
    tasks / executions: {
      href: "http://localhost:9393/dataflow/tasks/executions"
    },
    tasks / executions / name: {
      href: "http://localhost:9393/dataflow/tasks/executions{?name}",
      templated: true
    },
    tasks / executions / execution: {
      href: "http://localhost:9393/dataflow/tasks/executions/{id}",
      templated: true
    },
    jobs / executions: {
      href: "http://localhost:9393/dataflow/jobs/executions"
    },
    jobs / executions / name: {
[...]

PR #1770, opened 2021-09-30
Fix dynamic base path resolving

klopfdreh:
Addresses issue: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/747

The context path is now preserved from the initial call so that if `server.servlet.context-path` is set the ui automatically adopts it.

I did some initial tests, but would you be so kind and also test the functionality of the ui, so that there are no other issues I didn't figured out, yet.

As you can see in the following screenshots the "demoContext" base href is preserved while navigating.

This is a call against the dashboard uri:
dashboard_call

This is a call against the tasks uri:
task_call


klopfdreh:
There was an issue during the build with compodoc. The build was mentioning that I should file in a new Issue to report it. I noticed that the version was a bit outdated. I updated it from 1.0.4 to 1.1.15.

Locally I was able to fix this issue and tested it with `npm run doc`:
Bildschirmfoto 2021-10-04 um 18 18 14

jvalkeal:
I polished this PR a little when merged:

- Removed change to compodoc dep as dep updates should not get done with generic PR's
- Don't update lock file
- Removed use of APP_BASE_HREF which then made LeadingBaseHrefLocationStrategy defunct
- Restored path redirect from '' to 'apps'
- Fixed most of linting errors and left indentation errors in place as there's
  a know default conflict with prettier/eslint which needs to be fixed later
- We now get links like http://localhost:9393/scdf/dashboard/index.html#/apps
- Hard reload now works
- Generic polish

klopfdreh:
Hey @jvalkeal - I just tested the ui from master branch ( commit  5caf4a7f9e3cb3ee18b2c56195b05741c9e8864a ) with scdf server 2.9.0 and I can confirm that everything is working like a charm, now.

Refresh is working and also the initial redirection to Applications (apps)

Thanks for you help! 😃

Manu10744:
@klopfdreh 
I downloaded the `spring-cloud-dataflow-server-2.9.2.jar` from https://mavenrepository.com and modified the `application.yaml` as follows:

```yaml
server:
    port: 8337
    servlet:
        context-path: /scdf
```

Though, the dashboard still breaks at my end?

Console logs:
image

The JSON from `http://localhost:8337/scdf`
image

`http://localhost:8337/scdf/dashboard` only shows a frozen loading indicator and some errors in console:
image


The jar has `spring-cloud-dataflow-ui-3.2.2.jar` as dependency, so the fix is included, right? Have i done anything wrong??

klopfdreh:
I just had a look into the 3.2.2 tag and didn't noticed any of my changes - I guess they are not released, yet. @oodamien - I hope I didn't miss anything.

PR #1783, opened 2021-11-02
Fix some leftovers from dynamic base paths

klopfdreh:
Hey @jvalkeal - I just found some leftovers and also a not closed div tag in the logout-oauth-success.html

After I found those occurrences I did a full search for `href="/` but there are no others. I guess after this PR the change is finished.

### DIFF
--- a/ui/src/app/about/user/user.component.html
+++ b/ui/src/app/about/user/user.component.html
@@ -12,6 +12,6 @@
 
 <ng-template #logIn>
   <div>
-    <a href="/login" class="nav-link nav-icon-text"> Log In </a>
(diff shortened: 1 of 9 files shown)


klopfdreh:
Belongs to https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/747

klopfdreh:
I changed baseUrl to baseApiUrl, because otherwise the context-path is not used as it was: `/` compared to`<context-path>/` but with dashboard at the end which is wrong.

jvalkeal:
Thx, I just started to do some more work around security so this comes at a good time!

jvalkeal:
This was nice!. Merged per 2dc865cdfa50f973e8f68bd8bcb1fd02e64a4ee6

PR #1785, opened 2021-11-04
Update version label on Task Launch page

### LINKED ISSUE #1779: Default version is ignored when a previous launch of this task definition used another version.
Issue: Default version is ignored when a previous launch of this task definition used another version.
Steps to reproduce:

User launches task using a different version than the default.
User launches task again but this time selecting the default version enumerated at the top of selection list.
Expected:
The default version is utilzed when launching the task.

What Happens:
The version selected from the previous run is used.
This is because when the user specifies the default version after executing the task before with a another version, the UI does not send the version, thus data flow uses the previous version.   
In this case the UI should send the version if the previous execution was not the default.   

More information can be found here: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4724

oodamien:
Show a specific message on the app version field (Last launch version)
Resolves #1779

### DIFF
--- a/ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
+++ b/ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
@@ -28,6 +28,7 @@ import {PropertiesDialogComponent} from '../../../../flo/shared/properties/prope
 import {StreamAppPropertiesSource, StreamHead} from '../../../../flo/stream/properties/stream-properties-source';
 import {TaskPropertiesDialogComponent} from '../../../../flo/task/properties/task-properties-dialog-component';
 import {ValuedConfigurationMetadataProperty} from '../../../../shared/model/detailed-app.model';
+import {get} from 'lodash';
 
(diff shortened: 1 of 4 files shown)


cppwfs:
LGTM

ui/src/app/tasks-jobs/tasks/launch/builder/builder.component.ts
```diff
@@ -1118,6 +1120,14 @@ export class BuilderComponent implements OnInit, OnDestroy {
     }
   }
 
+  getLabelAppVersion(taskLaunchConfig, app) {
+    const lastVersion = get(taskLaunchConfig?.lastExecution?.deploymentProperties, `version.${app.name}`);
+    if (lastVersion && app.version !== lastVersion) {
+      return `Last launch (${lastVersion})`;
```
klopfdreh, line 1126: Maybe it would be nice to integrate Angular i18n instead of using fixed english Strings: https://angular.io/guide/i18n-overview

cppwfs:
Looks like the build is failing on this branch.  Can you take a look.

PR #1790, opened 2021-12-10
fix: Logos missing in about dialog

claudiahub:
- Solved rendering of icons inside 'About' modal


jvalkeal:
Perfect, gonna squash and polish commit message to add `fixes` which closes relevant issue.

PR #1816, opened 2022-02-09
fix: Update viewLog accessibility on task page

### LINKED ISSUE #1814: Button "View log" not available when task not finished
Moving this issue from Spring Cloud Deployers to here: https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/480
Unless there is an issue, we can allow users to view a log if they have an external-execution-id.   They don't have to wait for the task to be complete or in error to view.

oodamien:
Resolves #1814

### DIFF
--- a/ui/src/app/tasks-jobs/executions/execution/execution.component.ts
+++ b/ui/src/app/tasks-jobs/executions/execution/execution.component.ts
@@ -117,10 +117,7 @@ export class ExecutionComponent implements OnInit {
   }
 
   hasLog(): boolean {
-    return (
-      (this.execution.taskExecutionStatus === 'COMPLETE' || this.execution.taskExecutionStatus === 'ERROR') &&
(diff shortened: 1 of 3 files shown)


szopal:
I think that this change is not enough - you've changed only Task tab, but in this tab are visible Tasks on Last execution section which has status COMPLETED or ERROR, I think you should change also:
ui/src/app/tasks-jobs/jobs/execution/execution.component.ts

And for SpringBatch job:
ui/src/app/tasks-jobs/executions/execution/execution.component.ts


where is also used hasLog method with check status.

ilayaperumalg:
@szopal Thank you for the review and suggestion. We would be happy to welcome your PR contribution(s) based on your suggestions as well.

ilayaperumalg:
@oodamien Thank you Damien! 
@szopal A suggestion: You could add in-line comments instead of describing the files in the comments section.

PR #1873, opened 2022-07-04
fix: add missing translation

### LINKED ISSUE #1872: Translation issue in Job executions Steps table
Hey,

I just found another translation issue at the `Job executions` page. In the `STEPS` table there is the translation missing for the last column. `jobs.step.status` - I guess the label should be `Status` (this is also the translation for german `Status`)

image

oodamien:
Resolves #1872


klopfdreh:
Looks good to me! 👍

PR #1879, opened 2022-07-13
Prevent undefined on http task execution request

### LINKED ISSUE #1877: Bad request when there are no task executions
Hey everyone,

we currently have an issue when there are not task executions available for a task. The ui then try to make a request against `/undefined`. I guess this occurs because the last task execution could not be determined and the ui resolves `undefined` for it.

image

SCDF Server: 2.9.4
SCDF UI: 3.3.0-SNAPSHOT (current state of main branch)

I guess this could be prevent by just checking the last execution id not to be undefined.

claudiahub:
Resolves #1877

### DIFF
--- a/ui/src/app/shared/api/task.service.ts
+++ b/ui/src/app/shared/api/task.service.ts
@@ -156,7 +156,7 @@ export class TaskService {
 
   getExecution(executionId: string): Observable<TaskExecution | unknown> {
     const headers = HttpUtils.getDefaultHttpHeaders();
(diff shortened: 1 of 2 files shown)


klopfdreh:
Exactly what I thought the fix could look like. 👍

claudiahub:
Thanks @klopfdreh, good point!
I'd have applied that detail first but I was in hurry to fix this asap and I didn't noticed 😄

PR #1882, opened 2022-08-02
Fix translation for tasks execution cleanup

### LINKED ISSUE #1881: Translation issue in modal dialog of cleaning up task executions (modal body)
There is an issue when you clean up all task executions. As you can see here:

https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/main/ui/src/app/manage/tools/cleanup/cleanup.component.html#L10

and here

https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/main/ui/src/assets/i18n/en.json#L343

the property `tools.modal.cleanUp.removing` is missing - my assumption is that this could be translated with

EN: `Removing all entries.`
DE: `Entferne alle Einträge.`

claudiahub:
Resolves #1881


klopfdreh:
Looks good! 👍

Issue #1920, opened 2023-05-15
Theme dark highlight color should be adjusted


klopfdreh:
The highlight color of the dark theme is very bright. It should be adjusted to be more dark so that it doesn't hurt in eyes. 😀 

![image](https://github.com/spring-cloud/spring-cloud-dataflow-ui/assets/980773/8538f7f3-f011-431a-a984-515bd0e5903c)

SCDF Version 2.10.3 / Angular 15

PR #1922, opened 2023-05-17
fix: datagrid highlight color

klopfdreh:
fixes: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1920

The highlight of the datagrid-rows is much less intrusive, now.

Bildschirmfoto 2023-05-17 um 07 37 38


claudiahub (commented):
Thank you Tobias, let me know if you need the selector for the dark theme.
Other than that, LGTM.

klopfdreh:
I took me a while to find the type script files in which the variables are defined.

I thought it would be good to take the one which is responsible for the hover for the datagrid-row `--clr-datagrid-row-hover` and assign a good value for light and dark theme. The light theme looks like this, now:

Bildschirmfoto 2023-05-18 um 00 26 13

claudiahub (approved):

claudiahub:
Thanks @klopfdreh, I did some tests on different tables through Dataflow UI both in dark and light mode.
LGTM
image

oodamien (approved):
LGTM, thanks @klopfdreh

PR #1930, opened 2023-07-07
Fix duration label key

sayapeg:
When we select specific step execution from the job execution screen, on the left where the task execution details are shown, while showing the duration, the label Start Time is used instead of duration.
image


claudiahub (approved):
LGTM, thank you @sayapeg

Issue #1942, opened 2023-08-03
Cleanup task executions is not shown if "Streams" appFeature is disabled


klopfdreh:
The div in the component is placed in the wrong container. I am going to fix this.

PR #1943, opened 2023-08-03
fix: clean up should be available if streams are disabled

klopfdreh:
Fixes: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1942

`CLEAN UP TASK EXECUTIONS` has been moved into the container with `appFeature="tasks"`

Because this looks weird, as on the right side there would be two `app-view-card` and on the left only one I decided to make them all `clr-col-lg-12` to get the max width of the parent container.

Bildschirmfoto 2023-08-03 um 08 17 43


claudiahub (approved):
LGTM

PR #1997, opened 2024-04-09
Update task execution list

### LINKED ISSUE #1996: Update Task Execution list to use tasks/thinexecutions
The backend provides a new endpoint for retrieving pages of task executions without all the extra detail. 
The name of the link and url is `tasks/thinexecutions` and the resulting response has `_embedded.taskExecutionThinResourceList` instead of `_embedded.taskExecutionResourceList`. The objects in the page looks like TaskExecution but with some fields missing:
```TypeScript
class TaskExecutionThin {
    executionId: number;
    parentExecutionId: number;
    exitCode: number;
    taskName: string;
    startTime: DateTime;
    endTime: DateTime;
    exitMessage: string;
    externalExecutionId: string;
    taskExecutionStatus: string;
    errorMessage: string;
    schemaTarget: string;
    platformName: string;
    _links: any; 
}
```
`_links.self` for detail of TaskExecution when clicking on id
`_links['tasks/definitions']` for query when clicking on name.

claudiahub:
Resolves #1996

### DIFF
--- a/ui/src/app/shared/api/task.service.ts
+++ b/ui/src/app/shared/api/task.service.ts
@@ -44,14 +44,14 @@ export class TaskService {
       params = params.append('sort', `${sort},${order}`);
     }
     return this.httpClient
-      .get<any>(UrlUtilities.calculateBaseApiUrl() + 'tasks/definitions', {headers, params})
+      .get<any>(UrlUtilities.calculateBaseApiUrl() + 'tasks/executions', {headers, params})
       .pipe(map(TaskPage.parse), catchError(ErrorUtils.catchError));
   }
 
   getTask(name: string, manifest = false): Observable<Task | unknown> {
     const headers = HttpUtils.getDefaultHttpHeaders();
     return this.httpClient
-      .get<any>(UrlUtilities.calculateBaseApiUrl() + `tasks/definitions/${name}?manifest=${manifest}`, {headers})
+      .get<any>(UrlUtilities.calculateBaseApiUrl() + `tasks/executions/${name}?manifest=${manifest}`, {headers})
       .pipe(map(Task.parse), catchError(ErrorUtils.catchError));
   }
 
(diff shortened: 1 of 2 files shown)


oodamien (approved):

corneil (changes requested):
The thinexecutions output has `_embedded.taskExecutionThinResourceList` not `_embedded.taskExecutionResourceList`

corneil:
We also need a change in [task.service.ts](https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/21825fac0d3a7d84463e6574282878787d711b15/ui/src/app/shared/api/task.service.ts#L219)

ui/src/app/shared/api/task.service.ts
```diff
@@ -44,14 +44,14 @@ export class TaskService {
       params = params.append('sort', `${sort},${order}`);
     }
     return this.httpClient
-      .get<any>(UrlUtilities.calculateBaseApiUrl() + 'tasks/definitions', {headers, params})
+      .get<any>(UrlUtilities.calculateBaseApiUrl() + 'tasks/executions', {headers, params})
```
corneil, line 47: Why these changes? I was only expecting `tasks/executions` in `getExecutions` to become `tasks/thinexecutions`

corneil (changes requested):

claudiahub:
@corneil when reading the description on https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1996#issue-2232932015 (see the first comment)
I've read the mention to  `_links['tasks/definitions']`. 
Is this is for another purpose, correct ?

corneil:
> @corneil when reading the description on [#1996 (comment)](https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1996#issue-2232932015) (see the first comment) I've read the mention to `_links['tasks/definitions']`. Is this is for another purpose, correct ?

That link was added in case you wanted to use it to retrieve the detail. I believe the UI already has an approach in place which works and does the same thing. When people use the REST API and previously used `tasks/executions` and not use the faster `tasks/thinexecutions` they will want to retrieve the previous level of detail then using the link makes that easy.

ui/src/app/shared/api/task.service.ts
```diff
@@ -44,14 +44,14 @@ export class TaskService {
...
       .pipe(map(TaskPage.parse), catchError(ErrorUtils.catchError));
   }
 
   getTask(name: string, manifest = false): Observable<Task | unknown> {
     const headers = HttpUtils.getDefaultHttpHeaders();
     return this.httpClient
-      .get<any>(UrlUtilities.calculateBaseApiUrl() + `tasks/definitions/${name}?manifest=${manifest}`, {headers})
+      .get<any>(UrlUtilities.calculateBaseApiUrl() + `tasks/executions/${name}?manifest=${manifest}`, {headers})
```
corneil, line 54: Do these represent an old error that didn't surface?

oodamien:
Close for https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/1998

PR #1998, opened 2024-04-17
Update task execution list

### LINKED ISSUE #1996: Update Task Execution list to use tasks/thinexecutions

oodamien:
Resolves #1996

### DIFF
--- a/ui/src/app/app.module.ts
+++ b/ui/src/app/app.module.ts
@@ -30,7 +30,6 @@ import {AppsModule} from './apps/apps.module';
 import {UrlUtilities} from './url-utilities.service';
 import {TaskService} from './shared/api/task.service';
 import {JobService} from './shared/api/job.service';
-import {DevModule} from './dev/dev.module';
 
 @NgModule({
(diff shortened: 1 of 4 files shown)


corneil (approved):
LGTM

PR #2010, opened 2024-08-16
Use version rather than 'latest' in starter urls

onobc:
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/5897


onobc:
The [PR in the Microsite](https://github.com/spring-io/dataflow.spring.io/pull/524) with the new version URLs was merged and deployed into PROD. I have verified this works with Dataflow main.

PR #2029, opened 2025-01-07
Remove old badges from UI README page

cppwfs:
Badges for tools that are no longer in use are to be removed.


onobc (approved):
Thanks @cppwfs

